PR #10, opened 2016-08-02
New driver for Olimex MOD-LCD1x9

haraldg:
This time I also upated the Changelog and the CREDITS.
I hope this is how you want things to work. - If not, just tell me! :-)


willfe:
Looks good to me! I think you're right (in the comment you made when changing the CREDITS/ChangeLog) that referencing individual changes (and pulls) is going overboard. Your approach looks good, so it gets my vote :)

My apologies for attributing your work incorrectly. Thanks for fixing it!

I'll get this merged shortly.

PR #11, opened 2016-08-03
Allowing multibyte characters

haraldg:
Hi,

this PR is a bit more controversial than the earlier ones. It relates to code that has a lot of FIXMEs and whether the patch does the right thing depends on how we want to address the FIXMEs (if anybody ever cares).

Anyway, I think the change doesn't break anything that isn't already broken and allows many simple cases to just work.

In the long run we probably should have a strategy how to deal with character encodings including multibyte. For now this patch impements the "agnostic strategy" - pass anything to the display driver unmodified.


haraldg:
I forgot to mention: This patch didn't get any comments (positive or negative) when I posted it to the mailing list last year.

PR #15, opened 2016-08-05
Submitting a patch from the mailing list archive

haraldg:
It seems that Jörg can't be reached (not employed at pyramid.de anymore) so I'm creating this PR for the last patch he submitted to the mailinglist a few month ago.

### DIFF
--- a/server/drivers/glk.c
+++ b/server/drivers/glk.c
@@ -188,6 +188,11 @@ glk_init(Driver *drvthis)
 	p->width = 40; p->height = 8;
 	p->gpo_count = 1;
 	break;
+      case 0x20 :
+	p->model = "GLK12232-25-SM-USB";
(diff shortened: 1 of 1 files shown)


willfe:
Looks like this has some conflicts. Anybody want to take a stab at it?

haraldg:
Thats, odd. I did a clean fork from the master branch for this pull request. The only possible merge conflict I can think of, is the ChangeLog. I will update the PR once I'm back at my machine, but in the long run maybe we shouldn't update the ChangeLog from within PRs?

willfe:
Heh! Yeah, maybe we shouldn't update the changelog from PRs.

haraldg:
Ok, so I dropped the update of the change log from this PR. Looks like there are no conflicts now.

PR #20, opened 2016-09-26
Add support for Raspberry Pi 3

jerryr:
I've added some code to support the RPi3. Ideally we should be parsing the revision from /proc/cpuinfo using something like the code here: https://github.com/AndrewFromMelbourne/raspberry_pi_revision/blob/master/raspberry_pi_revision.c, but this works for now.


willfe:
Looks good, thanks!

PR #21, opened 2016-10-16
Update machine_OpenBSD.c

brendanhoran:
Fix make issues on OpenBSD 5.9+

"machine_OpenBSD.c:48:24: error: sys/dkstat.h: No such file or directory"

sys/dkstat.h is no longer part of OpenBSD 5.9
Remaining defines from this file can already be found in sys/sched.h


willfe:
Thanks for the patch!

PR #23, opened 2016-11-07
Minor tweaks to help the generation of the configuration GUI based on cme

dod38fr:
Hello

Here are some minor changes to `LCDd.conf` that may seem purely cosmetic from your point of view, but they do help the [generation](https://ddumont.wordpress.com/2011/07/03/generate-a-configuration-editor-from-a-config-template-file-with-perl-lcdproc-example/) of [LCDd configuration GUI](https://github.com/dod38fr/config-model/wiki/Managing-Lcdproc-configuration-with-cme).  The rationale of the changes are explained in each commit log.

Could you merge these changes so I don't need to maintain my own version of `LCDd.conf` ?

All the best

### DIFF
--- a/LCDd.conf
+++ b/LCDd.conf
@@ -811,7 +811,7 @@ Size=16x2
 [linux_input]
 
 # Select the input device to use [default: /dev/input/event0]
-Device=/dev/input/event0
+# Device=/dev/input/event0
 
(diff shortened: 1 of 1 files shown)


willfe:
Seems pretty reasonable, thanks for the patch!

PR #24, opened 2016-11-10
New hd44780 connection type using the linux sysfs gpio interface

haraldg:
This is yet an other patch from the mailing list archive. I did some minor cleanups, but as I don't have the HW to test this driver myself, I will wait for feedback before actually merging.

Also this PR doesn't include updates to the list of contributors, to avoid merge conflicts.


antoneliasson:
Nice work! I'll have a look this weekend. I have a few HD44780 wired to PCF8574T I2C GPIO expanders. There seems to be a kernel GPIO driver for this chip, although I haven't tested it myself:

https://github.com/torvalds/linux/blob/master/drivers/gpio/gpio-pcf857x.c
https://github.com/torvalds/linux/blob/master/Documentation/devicetree/bindings/gpio/gpio-pcf857x.txt

haraldg:
As it happens I just stumbled upon libugpio: https://github.com/mhei/libugpio - I will see how much code can be dropped from this PR if we use that.

antoneliasson:
Tested the patch, and it works (with some issues, see below)!

I'm using a Raspberry Pi 1 B running Raspbian with the latest kernel (4.4.26+). I have a PCF8574 based HD44780 i2c backpack which is described at [[1](https://arduino-info.wikispaces.com/LCD-Blue-I2C)] as "I2C LCD DISPLAY VERSION 1". Raspbian doesn't ship the gpio-pcf857x driver so I had to rebuild the kernel. After loading the module the i2c chip is initialized like this (i2c bus 1, addr 0x27):

```
$ echo pcf8574 0x27 | sudo tee /sys/bus/i2c/devices/i2c-1/new_device
```

The GPIO chip itself is then added under /sys/class/gpio/gpiochip504. So the GPIO base number is 504.

The documentation part of this patch says that:

> A configuration option for R/W pin of the display does not exist
>  since the driver always performs write operations and keeps the direction of
>  data pins as output. Therefore, the R/W pin <emphasis>must</emphasis> be
>  wired to GND. This is also necessary to prevent 5V logic appearing on
>  I/O pins which are not 5V tolerant.

This is stupid and irresponsible. Software works around hardware limitations, not the other way around. R/W is tied to a GPIO on many commercially available i2c backpacks and this driver should support that.

To make this driver work, I can manually export the R/W pin, set it to output low, unexport it and then run LCDd with the configuration described in the documentation of this patch. But this should be done automatically by the driver.

Another issue is that the driver does not unexport the GPIO pins on exit. Should the user wish to unload the GPIO kernel driver, the pins must first be manually unexported, otherwise rmmod will report the module as "in use".

IMHO, these are two major issues that should be solved before merging this patch.

dglaude:
I believe this code is not for those with I2C backpack nor does need
a PCF8574.
It might be more for those consuming 8(?) GPIO directly from the Pi to
directly to the HD44780.

antoneliasson:
It could be. This patch could easily replace the whole hd44780-i2c parts of lcdproc. It is also useful for I2C port expanders that have Linux kernel drivers but no lcdproc driver, such as the MCP23008.

haraldg:
Thanks for your testing and input!

About R/W pin: I agree that this driver should also be used for i2c port expanders.- Using the kernel drivers instead of shipping our own is the right way to do things, so I will update the PR in the way you described to support this mode of operation.

About the unexporting of the GPIOs on exit: I'm a bit surprised that this didn't work. The necessary code is present in gpio_HD44780_close(). So either the function wasn't called or the unexporting failed for some reason. How did you terminate LCDd?

The switch to libugpio looks good - maybe this smoothes some issues too. However I'm still battling autoconf to get things properly detected and built.

server/drivers/hd44780-gpio.c
```diff
@@ -0,0 +1,352 @@
...
+ *
+ * \param num   Number of the GPIO pin.
+ * \return      0 on success; -1 on error.
+ */
+static int
+unexport_gpio(int num)
+{
+	FILE *f = fopen("/sys/class/gpio/unexport", "w");
+
+	if (f == NULL)
```
antoneliasson: Add a perror() here. In my testing I get _Permission denied_ while lcdproc tries to unexport every GPIO pin during shutdown.

server/drivers/hd44780-gpio.c
```diff
@@ -0,0 +1,352 @@
...
+	int fd;
+
+	if (pin->number < 0) {
+		report(RPT_ERR, "setup_gpio_pin: pin_%s must be set to a valid value", pin->name);
+		return -1;
+	}
+
+	if (export_gpio(pin->number) != 0) {
+		report(RPT_ERR, "setup_gpio_pin: unable to export GPIO %d: %s",
+		       pin->number, strerror(errno));
```
antoneliasson: Should _include string.h_ for strerror to be found.

antoneliasson (changes requested):
By printf debugging I found that unexport_gpio() gets called a number of times during shutdown, but fails to open /sys/class/gpio/unexport with Permission denied errors. No idea why :/

antoneliasson:
Regarding termination: I normally run LCDd in the foreground during testing and exit with C-c, but I've also tried sending SIGTERM with the same result (as expected).

antoneliasson:
The driver termination procedures seem to run in a signal handler context by looking at server/main.c. There are many things you are not allowed to do in a signal handler, but you should be able to open files safely according to signal(7). I tried rewriting unexport_gpio to use the open(2) syscall directly instead of fopen(3), but had the same result. Maybe sysfs file are magic and have more restrictions.

I think the standard procedure for writing signal handlers is to only set a global flag that worker threads check periodically and if they found it to be set, they kill themselves cleanly. But that's a bigger change in LCDproc that I'm not willing to undertake myself at this point.

haraldg:
Thanks for your investigation! A strange issue, I will complete the transformation to libugpio first - maybe this magically fixes things. Then I will come back to debugging this - I guess I should be able to reproduce this even without the display HW.

haraldg:
Just updated the PR as per your review comments. Can you check, if this fixes all issues? You now need [libugpio ](https://github.com/mhei/libugpio) as new dependency. Does this fix the unexport problem? I'm going to do a few tests on my olinuxino board - however without a real display to drive.

Can somebody with autoconf knowledge have a look it what I did is sane?

haraldg:
Ok, actually figured that unexport problem out: LCDd is dropping root privileges after setting up everything! I didn't know it does that - this explains everything. I think the only way to fix this is not dropping root. I won't enforce this in the code though: It should be the users choice if they want to run as root or live with claimed GPIOs after exit. However I will add a note to the userguide.

docs/lcdproc-user/drivers/hd44780.docbook
```diff
@@ -3209,9 +3209,12 @@ controller.</para>
...
 
 <warning>
-  <para>A configuration option for R/W pin of the display does not exist
-  since the driver always performs write operations and keeps the direction of
-  data pins as output. Therefore, the R/W pin <emphasis>must</emphasis> be
+  <para>The configuration option for R/W pin of the display is optional.
+  If the pin is configured, it is always held in low state since
+  the driver always performs write operations and keeps the direction of
+  data pins as output. This is what you need for i2c port expanders.
+  If you ommit configuration of R/W pin (maybe you are manually wiring
```
antoneliasson, line 3216: Typo in 'omit'.

server/drivers/hd44780-gpio.c
```diff
@@ -21,6 +21,7 @@
 #include <fcntl.h>
 #include <stdio.h>
 #include <stdlib.h>
+#include <string.h>
```
antoneliasson, line 24: I'd squash this change into the previous commit, since it hasn't been merged yet.

acinclude.m4
```diff
@@ -216,11 +216,12 @@ dnl			else
...
 			HD44780_DRIVERS="hd44780-hd44780-serial.o hd44780-hd44780-lis2.o hd44780-hd44780-usblcd.o"
-			case $host in
-			*-*-linux*)
+			AC_CHECK_LIB(ugpio, main,[
 				HD44780_DRIVERS="$HD44780_DRIVERS hd44780-hd44780-gpio.o"
-				;;
-			esac
+				LIBUGPIO="-lugpio"
+			],[
+				AC_MSG_WARN([Could not find libugpio])
```
antoneliasson, line 223: Maybe append ", not building hd44780-gpio driver" to the warning string?

antoneliasson (approved):
Tested successfully. It does make this driver a bit more difficult to build (and possibly run) since libugpio is not packaged in any major distribution. I'm not sure how to handle that in the next lcdproc release. One option is of course to bundle libugpio.

I can confirm that unexporting works when lcdproc is run with privilege dropping disabled. Nice catch!

antoneliasson:
In other news, I found out that the sysfs GPIO interface was deprecated in Linux very recently. See https://www.youtube.com/watch?v=lQRCDl0tFiQ&list=PLbzoR-pLrL6pRFP6SOywVJWdEHlmQE51q&index=83 from around 37 minutes. The "new" way to access GPIOs from user space (if you really don't want to write a kernel device driver) is through a character device. A major benefit of this approach compared to the old is that exports will be automatically cleaned up on program exit (even on crashes).

It doesn't really affect this project for several years though, since the change is only available in Linux 4.8 or 4.9 and up.

haraldg:
Thank you for your mindful comments.

About the new way to access GPIOs: Thats a reason why I'm happy I ported the driver to libugpio - I expect it will track changing kernel ABI for us. Yes, it's not yet widely included into many distributions, but I hope disitribution maintainers of lcdproc will package it. @dod38fr do you have any opinion about this? Are there any non-debian based linux distributions that ship lcdproc?

The embarassing thing about the privilege dropping issue is: It was my first suspect, but when grepping for it, I had a false negative...

dod38fr:
@haraldg : I can package `libugpio` for Debian. 

lcdproc is also [packaged for Fedora and Mageia](http://www.rpmfind.net/linux/rpm2html/search.php?query=lcdproc&submit=Search+...&system=&arch=)

All the best

haraldg:
Ouch, this PR was still missing final updates to address the last review comments. I guess I should have added some prefix like "WIP:" to the subject line...

I'll just push a commit with final updates directly. Sorry for the noise.

antoneliasson:
> About the new way to access GPIOs: Thats a reason why I'm happy I ported the driver to libugpio - I expect it will track changing kernel ABI for us.

I wouldn't expect that in this case. The reason is that the old sysfs interface exposes magic GPIO numbers to userspace (and libugpio in turn expose them to the user). These numbers are platform dependent and unstable. For example, the GPIOs on the port expander on my system are eight consecutive numbers starting at 504 for no particular reason. Assuming libugpio gets updated to the upcoming kernel API, it would have to emulate these numbers somehow to avoid breakage.

But that's a later problem. Great job with this patch @haraldg and @fatih-asici !

PR #25, opened 2016-11-20
Update documentation to replace references from CVS to Git

coliss86:
Here is an update of the documentation to replace references from CVS to Git.

As I don't know how you want to update the release process, the file `docs/lcdproc-dev/releasing.docbook` has not been modified.

### DIFF
--- a/docs/lcdproc-user/how-to-obtain.docbook
+++ b/docs/lcdproc-user/how-to-obtain.docbook
@@ -25,8 +25,8 @@ around on the internet:
   <term>Older releases</term>
   <listitem>
     <para>
-    Older releases of LCDproc can be found on the Sourceforge project page at
-    <ulink url="http://sourceforge.net/projects/lcdproc/files/lcdproc/"></ulink>.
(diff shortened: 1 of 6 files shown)


docs/lcdproc-user/how-to-obtain.docbook
```diff
@@ -25,8 +25,8 @@ around on the internet:
   <term>Older releases</term>
   <listitem>
     <para>
-    Older releases of LCDproc can be found on the Sourceforge project page at
-    <ulink url="http://sourceforge.net/projects/lcdproc/files/lcdproc/"></ulink>.
+    Older releases of LCDproc can be found on the GitHub project page at
+    <ulink url="https://github.com/lcdproc/lcdproc"></ulink>.
     </para>
```
haraldg, line 30: This is not true at the moment. I'm not sure what should be written here instead, though.

willfe: I'll look into how we can host tarballs of older releases on Github so we can leave the statement there (pointing to Github).

PR #29, opened 2016-12-16
NoritakeVFD: Fix # Of Custom Characters

hoblins:
The display the driver was tested with, CU40026SCPB-T20A, supports 8 custom characters according to datasheet. Page 2/4 "Escape Commands", "8 User Defineable Characters (UDC)".
I tested the patch with with the CU20045SCPB-T31A which also supports at least 8 custom characters.
As I'm not sure about external links, please search for PDFs of the datasheets if required. They are easy to find.
So big numbers can be displayed in nicer way, see adv_bignum.c.

### DIFF
--- a/server/drivers/NoritakeVFD.c
+++ b/server/drivers/NoritakeVFD.c
@@ -66,7 +66,7 @@
 
 
 /* Constants for userdefchar_mode */
-#define NUM_CCs		2 /* max. number of custom characters */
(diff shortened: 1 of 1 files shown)


haraldg:
Looks like even the CU20045SCPB-T28A for which this driver has originally been written supports 16 custom characters. If there are displays with different number of custom characters out there, maybe this should be a dynamic setting though. Anybody likes to work an this?

PR #30, opened 2016-12-16
NoritakeVFD Patch Cursor

hoblins:
The positioning of the cursor was not working correctly. Beside other fixes the rows and columns were inverted in the initial code.
Tested with CU20045SCPB-T31A.

### DIFF
--- a/server/drivers/NoritakeVFD.c
+++ b/server/drivers/NoritakeVFD.c
@@ -412,7 +412,7 @@ NoritakeVFD_chr (Driver *drvthis, int x, int y, char c)
  * \param string   String that gets written.
  */
 MODULE_EXPORT void
-NoritakeVFD_string (Driver *drvthis, int x, int y, const char string[])
+NoritakeVFD_string (Driver *drvthis, int x, int y, unsigned char string[])
 {
 	PrivateData *p = drvthis->private_data;
 	int i;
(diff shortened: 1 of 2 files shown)


server/drivers/NoritakeVFD.c
```diff
@@ -412,7 +412,7 @@ NoritakeVFD_chr (Driver *drvthis, int x, int y, char c)
  * \param string   String that gets written.
  */
 MODULE_EXPORT void
-NoritakeVFD_string (Driver *drvthis, int x, int y, const char string[])
+NoritakeVFD_string (Driver *drvthis, int x, int y, unsigned char string[])
 {
```
haraldg, line 416: It should be "const unsigned char" at least. However the API documentation quite clearly requires "const char" - maybe this is a bad idea. But then we should change this properly.

haraldg:
@hoblins Are you going to update or comment this PR?

I want to merge this PR before I tag the release and would edit this patch myself if you are not working on it.

server/drivers/NoritakeVFD.c
hoblins: I just took this change from another driver.
But I can update the commit and test the corrections by the end of next week.

hoblins:
Made changes remarked by haraldg and tested with CU20045SCPB-T31A.

PR #32, opened 2016-12-17
server/drivers: Make report() available globally

haraldg:
This PR really simplifies the API a lot. If this works, maybe we should do the same with the other function pointers passed via driver data as well. However the report() function is the one that was really annoying.

Can anybody think of a situation where this might break? Can anybody test this with other compilers then gcc or on an other OS then linux? (OS shouldn't matter I think.)

Is somebody unhappy, if I merge this before the release happens?


haraldg:
Pushed equivalent

PR #46, opened 2017-01-08
Small change to README.md and server/drivers/adv_bignum.c

emteejay:


haraldg:
Hi, I don't see the point of the slim variant and I don't like that your patch makes this the new default. (And without a way to change this other then editing the source.) Unless you update this PR I will probably just pick up the corrections and close it.

PR #47, opened 2017-01-11
dm140: add driver

awiouy:
Hello,
The dm140 driver exists as a patch to LibreELEC/OpenELEC.
It would however be profitable to merge it upstream.
Thank you for your attention and support.

PR #48, opened 2017-01-11
dm140 drivers

CvH:
First I am not the author of this patch, we carry it at OpenELEC btw now in LibreELEC for a long time and it works. I have no idea from where the patch comes from `lcdproc-0.5.6-dm140_henlar_v0.2.patch` was added 4 years ago. 

I have **only** fixed the diff against current master.


CvH:
sorry #47

PR #55, opened 2017-02-07
Yard2lcd

YARD2:
Update to avoid conflicts with -wall


server/drivers/yard2LCD.c
```diff
@@ -305,7 +308,7 @@ yard_init(Driver *drvthis)
 	}
 	
 	//Get the config from yard2srvd. Config is not in LCDd.conf !
-	sprintf(Recbuffer,"LCDPROC\0");
+	sprintf(Recbuffer,"LCDPROC\0"); // \0 is needed for YARD daemon
 	byteCnt = write(p->fd,Recbuffer, strlen(Recbuffer));
```
haraldg: I don't think this code works as you expect. (If I guess correctly what you want it to do.)
There won't be two zeros in `Recbuffer`

pprindeville: Yeah, you'd need:

```
sprintf(Recbuffer, "LCDPROC%c%c", 0, 0);
```

but please use `snprintf()` instead.  `sprintf()` is deprecated.

PR #60, opened 2017-02-12
musl: add headers for clean compilation with MUSL libraries

pprindeville:
Without this patch, compilation under MUSL 1.1.16 generates a lot of warnings about the wrong versions of headers being included.


haraldg:
I can't reproduce any related warnings with MUSL 1.1.15 - I think the issues are already fixed in the master branch (and 0.5.8) by other means. I will check against MUSL 1.1.16 if you insist. Maybe you can quote the warnings here?

haraldg:
Closing as obsolete.

PR #62, opened 2017-02-12
CwLnx: this function isn't defined except conditionally

pprindeville:
So it's shouldn't be forward declared except conditionally as well.


pprindeville:
This should be easy enough to confirm by inspection...

Later in the file you have:

```
#if defined(SEAMLESS_HBARS)
static void
CwLnx_set_char_unrestricted(Driver *drvthis, int n, unsigned char *dat)
{
    PrivateData *p = drvthis->private_data;
```

...

haraldg:
Sure, actually I'm taking the opportunity to just get rid of SEAMLESS_HBARS once and for all. This should be a per driver thing, not a compile time option for the server...

This is only waiting for a slot of my time to write code.

PR #63, opened 2017-02-12
t6963: move inline declarations into same module as function

pprindeville:
Static inlines need to be declared in the same module as they're defined.

### DIFF
--- a/server/drivers/t6963_low.c
+++ b/server/drivers/t6963_low.c
@@ -43,6 +43,9 @@
 #define T_DATA	0x00		/* ~INIT didn't work here */
 
 
+inline void
+t6963_low_send(T6963_port *p, u8 type, u8 byte);
+
 /**
  * Acquires access to parallel port and initializes timing. The parallel port
  * must be an I/O-address between 0x200 and 0x400.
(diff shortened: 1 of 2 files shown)


haraldg:
I don't understand the point of this patch. What's this supposed to fix?

pprindeville:
Without it, I get:

```
t6963_low.c:207:2: warning: 'port_out' is static but used in inline function 't6963_low_send' which is not static
  port_out(T6963_CONTROL_PORT(p->port), portcontrol ^ OUTMASK);
  ^
t6963_low.c:205:3: warning: 'timing_uPause' is static but used in inline function 't6963_low_send' which is not static
   timing_uPause(1);
   ^
t6963_low.c:202:2: warning: 'port_out' is static but used in inline function 't6963_low_send' which is not static
  port_out(T6963_CONTROL_PORT(p->port), portcontrol ^ OUTMASK);
  ^
t6963_low.c:200:2: warning: 'port_out' is static but used in inline function 't6963_low_send' which is not static
  port_out(T6963_DATA_PORT(p->port), byte);
  ^
t6963_low.c:199:2: warning: 'port_out' is static but used in inline function 't6963_low_send' which is not static
  port_out(T6963_CONTROL_PORT(p->port), portcontrol ^ OUTMASK);
  ^
```

PR #64, opened 2017-02-12
yard2LCD: various cleanup too numerous to fit in one line

pprindeville:
Don't declare functions that don't exist.

Don't use sprintf()'s it's unsafe, and don't embed \0 into C-strings.

Don't define functions with no return type.


haraldg:
The author of the yard2LCD driver has already submitted a PR #55 to address these issues. I'm not entirely happy yet, so I haven't merged it yet. Closing this as duplicate.

PR #66, opened 2017-02-12
configure: detect presence of nanosleep

pprindeville:
And if nothing else, annotate its existence in config.h which might be useful for troubleshooting cross-build issues.


haraldg:
I decided against this one:
* it doesn't really add anything and we will notice easily enough if somebody ports lcdproc to where there is no nanosleep()
* the code you propose to change is for the benefit of solaris, which I can't test. (It looks ok, but you never know)

pprindeville:
Not sure what there is to test.  It looks for the function (via linkage, not run-time execution) being found in a simple templatized program.  If that fails (the 3rd arg of `AC_CHECK_FUNC()`) then this alternate test is done.

More to the point, this is exactly how `connect()` and `inet_aton()` are handled on lines 86 and 87.

Also, if we have no means to test Solaris, then the logical conclusion of that is that any change we make could potentially break Solaris... causing development to come to a grinding halt.

Or we could take Solaris off life support, and say that if anyone cares enough to have it be maintained then they can volunteer to test fixes (or grant one of us access to a machine where we can at least do builds).

Unfortunate that Travis doesn't support Solaris.

PR #67, opened 2017-02-12
sdeclcd: don't rely on STRB bit being 1/0 when manipulating backlighting

pprindeville:
This code works on the lucky coincidence that backlighting is controlled
by the `STRB` bit whose value is 1, and that this function is called with
a 0 or 1 value to reset/set backlighting.

If I called `sdec_write(..., 99)` that would also be a valid way of
turning on the backlight, since that argument is ultimately zero or
non-zero in how it's interpreted.  But that would then mask in the
wrong value into the control register.

So, test backlight for being non-zero, but then assign the right bit
(which happens to be 1 or 0) into the variable we then combine to
generate the correct register value.

If we were running on different hardware were the 2 bit controlled
the backlighting, obvious this wouldn't work... so it's a lucky
coincidence that it does.

### DIFF
--- a/server/drivers/sdeclcd.c
+++ b/server/drivers/sdeclcd.c
@@ -51,8 +51,9 @@
  * mapping between the parallel port Control register and the
  * LCD control lines, as follows: (see lpt-port.h for defs)
  */
-#define SDEC_BACKLIGHT	STRB	/* Strobe, bit 0			*/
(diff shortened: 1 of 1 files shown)


pprindeville:
Redux per François' email today.

fmertz:
The proposed code change makes the backlight logic easier to understand. As the original author of this driver code, I am OK with merging this Pull Request into the main project code base.

PR #82, opened 2017-05-06
Hd44780 ezio

fmertz:
I wanted to propose a code change to add support for yet another variation of the HD44780 driver: the Portwell EZIO-100 and EZIO-300 devices. These devices come built-in some firewall appliances and have some popularity in the pfSense community.

Portwell put a serial interface "in front" of the LCD controller. The serial device has a couple of differences compared to other serial interfaces already implemented. The initialization sequence is "repeated", and a command has to be sent in order to receive the keypad status.

A concern is that I only own this particular EZIO-100 device, so that is the only one I can test. I therefore tried to minimize the possibility of accidentally breaking support for other devices by updating the main struct in the serial code.

The documentation is [here](http://drivers.portwell.com/CA_Manual/).

Thanks for considering this request for inclusion into the project. I will do a follow-up pull request for the updated documentation.

### DIFF
--- a/server/drivers/hd44780-serial.c
+++ b/server/drivers/hd44780-serial.c
@@ -6,7 +6,7 @@
  * and are connected to a serial port using some microcontroller. It supports
  * protocols using escape sequences to trigger commands, backlight or keys.
  *
- * Currently supported are: \c picanlcd, \c lcdserializer, \c los-panel,
+ * Currently supported are: \c ezio, \c picanlcd, \c lcdserializer, \c los-panel,
  * \c vdr-lcd, \c vdr-wakeup, and \c pertelian.
  */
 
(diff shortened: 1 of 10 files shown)


fmertz:
Build can be fixed with PR #83

server/drivers/hd44780-serial.c
```diff
@@ -362,16 +382,34 @@ serial_HD44780_scankeypad(PrivateData *p)
...
+						case 0xBD:
+							return 0x34;	/* KeyMatrix_4_3=Down   */
+						case 0x4E:
+						case 0xB7:
+							return 0x44;	/* KeyMAtrix_4_4=Escape */
+						/* No key 0x4F/0xBF or more than one key */ 
+						default:
+							return 0;
+					}
 				}
```
haraldg: Coding style want's the case statement on the same indentation level as the switch statement.

haraldg (changes requested):
Looks good to me, but I will only merge after Documentation has been added to the PR.

In the future I'd prefer smaller commits embodying logical changes, but no need to change this now.

server/drivers/hd44780-serial.c
fmertz: Another commit was provided to better comply with the coding style

PR #85, opened 2017-05-30
Fixes for FreeBSD 11

fmertz:
A couple of things come up while compiling under FreeBSD 11 AMD64:

- if HAVE_DEV_IICBUS_IIC_H is defined, we need to include the actual file because the code uses a definition from it. I have no hardware to actually test this out...

- clang issues a warning for **p = '\0';**. I corrected the warning. Second opinion welcome.


server/commands/screen_commands.c
```diff
@@ -559,7 +559,7 @@ screen_del_key_func(Client *c, int argc, char **argv)
 				*to++ = *from++;
 			}
 		}
-		to = '\0';	/* terminates the new keys string...*/
+		*to = '\0';	/* terminates the new keys string...*/
 	}
```
haraldg: I have looked at this code for half an hour now and not yet figured out what it's supposed to do. It predated git, the feature is undocumented, unused in our code base and quite clearly never worked. I say: Let's just rip out this function - nobody is going to miss it.

server/drivers/hd44780-i2c.c
```diff
@@ -85,6 +85,10 @@
 
 #include "i2c.h"
 
+#ifdef HAVE_DEV_IICBUS_IIC_H
+#include <dev/iicbus/iic.h>
+#endif
+
 // Generally, any function that accesses the LCD control lines needs to be
```
haraldg: AFAICS the code depending on this header file is stale, so lets remove the code instead of including the header.

haraldg (changes requested):
See inline comments.

If you had multiple commits, one for each logical change, in your PRs, it would be easier to cherry pick good stuff (or revert it, if it turns out to be a bad idea later).

server/commands/screen_commands.c
fmertz: The code has a logic to it, and I believe the proposed fix would make it work, but I agree that in the larger context, the code is basically unreachable. Best I can figure, the lookup has had the commands in front of this code commented out since at least 2002. I will update the pull request with a cleanup instead.

server/drivers/hd44780-i2c.c
fmertz: Agreed, more cleanup!

haraldg:
Looks good to me, only the changelog lines don't match the changes anymore ...

PR #89, opened 2017-09-05
Addded backlight change support for hd44780 compatible PT6314 VFD driver

trex2000:
Could you add this modification to support the backlight change for the HD44780 compatible PT6314 VFD driver?


server/drivers/hd44780-i2c.c
```diff
@@ -317,9 +317,30 @@ i2c_HD44780_senddata(PrivateData *p, unsigned char displayID, unsigned char flag
  */
 void i2c_HD44780_backlight(PrivateData *p, unsigned char state)
 {
+	static unsigned char old_state=0;
```
haraldg, line 323: Is this a safe assumption?

server/drivers/hd44780-i2c.c
```diff
@@ -317,9 +317,30 @@ i2c_HD44780_senddata(PrivateData *p, unsigned char displayID, unsigned char flag
...
 		p->backlight_bit = ((!p->have_backlight||state) ? 0 : p->i2c_line_BL);
 	else // Inverted backlight - npn transistor
 		p->backlight_bit = ((p->have_backlight && state) ? p->i2c_line_BL : 0);
 	i2c_out(p, p->backlight_bit);
+	if (state!=old_state)
+	{
+		    if (state == BACKLIGHT_ON)
+		    {
+    			/*100% brightness*/
+			p->hd44780_functions->senddata(p, 0, RS_INSTR, 0x28 );
```
haraldg: Maybe the constant should be defined somewhere ...

server/drivers/hd44780-i2c.c
```diff
@@ -317,9 +317,30 @@ i2c_HD44780_senddata(PrivateData *p, unsigned char displayID, unsigned char flag
...
+			p->hd44780_functions->uPause(p, 150);
+		    }
+		    else
+		    {
+    			/*25% brightness*/
+			p->hd44780_functions->senddata(p, 0, RS_INSTR, 0x2B );
+			p->hd44780_functions->uPause(p, 150);
+		    }
```
haraldg: The call to uPause is duplicate and should be moved out of the if-else block.
Actually the code could be simplified even further like
```
p->hd44780_functions->senddata(p, 0, RS_INSTR, state == BACKLIGHT_ON ? 0x28 : 0x2B);
```

haraldg (changes requested):
Thanks for your contribution. I have added a number of comments about the code inline. However there are a number of more general issues to be addressed:

The commit itself:
Please configure your git to include a real e-mail Address instead of root@something (BTW working as root is a bad idea anyway) and probably also your real name. (The latter is recommended but not required.)

I had a hard time figuring out what your change is intended to do. A commit message explaining the idea of the change would help a lot.

Changing the permissions on autogen.sh is an unrelated change and should be a separate commit, though frankly I don't quite see the point of it.

About the change:
The rpi and i2c connection types have been obsoleted by the gpio connection type. I won't accept new features for them unless the same feature is also added to gpio.

I can only guess, but it seems that your code is about a new display controller that can control the backlight internally instead of relying on an external pin. If so, then it is not clear that adding the new code unconditional is not going to cause regressions on the other controllers. At the very least your change causes an additional delay for everyone. How common is the instruction you are using in other controllers?

Since your code only uses controller instructions but no details of the connection type, maybe it would be best to add your code to the hd44780 common driver and add a configuration setting to decide whether common or connection type code is used for backlight handling ...

Finally please try to match the style of surrounding code regarding whitespace and indentation.

HTH,
Harald

server/drivers/hd44780-i2c.c
trex2000: I agree with you.

server/drivers/hd44780-i2c.c
trex2000: It was added to update the register only in case of a change. It's a safe assumption, as this would cause one additional update, at worst.

haraldg: I'm not concerned about extra updates, but about suppressing necessary updates. Is it guaranteed that upon initialization the device will be in state 0? Why not set old_state to some invalid value to force setting the backlight at least once?

trex2000:
So, I will address the issues mentioned above and will also do the update for the gpio. 
Will also update git with my real name and valid email address. I rarely do git commits (I hate git anyway) so I've slipped over this.
Regarding the modification, I own you indeed some explanations.
There is a VFD display driver, called PT6314, that is completely compatible with the HD44780, widely used for LCDs, but it has some additional registers, compared to HD44780.
Here is the datasheet:
https://www.newhavendisplay.com/app_notes/PT6314.pdf
This one is used in this LCD board that I own and use it for like 2 years now. First it was used with a rapsberry PI using GPIO, now I've switched to Odroid C2 and use it with an I2C backpack:
http://www.artronic.com.pl/o_produkcie.php?id=1101?
Here you don't have the backlight as a seprate input pin, as the backlight of the VFD is changed by changing the duty cycle of the PWM signal that drives the 2 transversal heating filaments of the VFD display.
Therefore, you can change, as with most VFD displays  between 4 predefined brightness values: 25%, 50%, 75%, 100%.
You can achieve this by modifying BR0 and BR1 in the function set register (see chapter 6.6. page 26 of datasheet).
Now, I see an improvement here, that could be done, to read out the register and mask out the rest of the bits and modify only BR0 and BR1. The rest is overwritten anyway cyclically, as I've seen, but will be safer anyway. It's good to have somebody review the code.
I've tested it with an normal LCD with HD44780, and seems to ignore this, so it's not in issue to send this also with HD44780.
But to be elegant, I could add a new parameter in the config file, to check if there is an VFD used, and send this only in that case. But here I need your help then :)

haraldg:
On 06.09.2017 11:02, trex2000 wrote:
> So, I will address the issues mentioned above and will also do the
> update for the gpio.
> Will also update git with my real name and valid email address. I
> rarely do git commits (I hate git anyway) so I've slipped over this.
> Regarding the modification, I own you indeed some explanations.
> There is a VFD display driver, called PT6314, that is completely
> compatible with the HD44780, widely used for LCDs, but it has some
> additional registers, compared to HD44780.
> Here is the datasheet:
> https://www.newhavendisplay.com/app_notes/PT6314.pdf

Thanks for the explanation. Most of this should be be added to the
commit message (or even the documentation of LCDd).

In this case I suggest not bothering about individual connection
types, but modifying the core driver instead. Ideally in a way that
is extensible to full PT6314 support in the future.

As a first step I suggest recursively grepping the whole source
tree case-insensitive for PT6314 to check for any prior work in this
direction. (I'm not at a computer with the source ATM.)

> This one is used in this LCD board that I own and use it for like 2
> years now. First it was used with a rapsberry PI using GPIO, now I've
> switched to Odroid C2 and use it with an I2C backpack:
> http://www.artronic.com.pl/o_produkcie.php?id=1101?
> Here you don't have the backlight as a seprate input pin, as the
> backlight of the VFD is changed by changing the duty cycle of the PWM
> signal that drives the 2 transversal heating filaments of the VFD
> display.
> Therefore, you can change, as with most VFD displays  between 4
> predefined brightness values: 25%, 50%, 75%, 100%.
> You can achieve this by modifying BR0 and BR1 in the function set
> register (see chapter 6.6. page 26 of datasheet).
> Now, I see an improvement here, that could be done, to read out the
> register and mask out the rest of the bits and modify only BR0 and
> BR1. The rest is overwritten anyway cyclically, as I've seen, but 
> will
> be safer anyway. It's good to have somebody review the code.
> I've tested it with an normal LCD with HD44780, and seems to ignore
> this, so it's not in issue to send this also with HD44780.

Well, there are dozens of clones out there, some of which extend the
registers in their own ways like the PT6314 apparently does. I think
we should be careful here.

> But to be elegant, I could add a new parameter in the config file, to
> check if there is an VFD used, and send this only in that case. But
> here I need your help then :)

Sure, how can I help?

I think there could be either a boolean flag telling the driver that
it actually is talking to a PT6314 or a more general extension 
parameter,
with PT6314 the first extension being implemented.

Either way the hd44780 init function would check this information and
change the function pointer p->hd44780_functions->backlight 
accordingly.
The hd44780 core driver would provide a hd44780_pt6314_backlight()
function for this purpose.

Note, that I don't have a HD44780 myself and am not familiar with this
driver, so all my remarks are just kind of general advice, that might
be wrong sometimes.

HTH,
Harald

-- 
If you want to support my work:
see http://friends.ccbib.org/harald/supporting/
or donate via peercoin to P98LRdhit3gZbHDBe7ta5jtXrMJUms4p7w
or CLAM xASPBtezLNqj4cUe8MT5nZjthRSEjrRQXN

trex2000:
I've just committed yesterday the updates for the PT6314 driver, but I need some input as currently I have only an vfd connected through an I2C backpack. 
The code seems to be quite similar for the rpi, gipio, eth and ftdi connections, so i've performed also the updates for these, but It needs to be tested. 
We can discuss the changes if you'd like.

server/drivers/hd44780-4bit.c
```diff
@@ -101,6 +101,10 @@ unsigned char lcdstat_HD44780_readkeypad(PrivateData *p, unsigned int YData);
 
 static const unsigned char EnMask[] = { EN1, EN2, EN3, STRB, LF, INIT, SEL };
 
+/*function set buffer to store during init*/
+static unsigned char bufferFUNCSET = 0;
+
```
haraldg: The LCDd driver API doesn't allow for static variables like this. You need to add this to the driver private data instead.

server/drivers/hd44780-4bit.c
```diff
@@ -262,9 +268,40 @@ lcdstat_HD44780_senddata(PrivateData *p, unsigned char displayID, unsigned char
...
 	port_out(p->port, p->backlight_bit);
+
+	if (p->isVFDDisplay)
+	{
+		if (state!=old_state)
+		{
+			//for PT6314 VFD driver call init function to update BR0-BR1 registers for backlight level
+			if (p->backlightstate)
+			{
+				brightnessLevelVFD = 0x03u; //100% brightness
```
haraldg: Since you define a macro for this constant in the headerfile, please use it. Numeric constants are always a bit hard to track in mind when reading the code.

server/drivers/hd44780-4bit.c
```diff
@@ -262,9 +268,40 @@ lcdstat_HD44780_senddata(PrivateData *p, unsigned char displayID, unsigned char
...
+	unsigned char brightnessLevelVFD = 0;
+
 	p->backlight_bit = ((!p->have_backlight||state)?0:BL);
 
 	port_out(p->port, p->backlight_bit);
+
+	if (p->isVFDDisplay)
+	{
+		if (state!=old_state)
+		{
```
haraldg: No need for nested if statements.
`if (p->isVFDDisplay && state != old_state)` is easier to read I think.

server/drivers/hd44780-4bit.c
```diff
@@ -262,9 +268,40 @@ lcdstat_HD44780_senddata(PrivateData *p, unsigned char displayID, unsigned char
...
+			}
+			else
+			{
+				brightnessLevelVFD = 0x01u;//25% brightness
+			}
+			/*update brightness level of VFD*/
+			p->hd44780_functions->senddata(p, 0, RS_INSTR, bufferFUNCSET | brightnessLevelVFD);
+			p->hd44780_functions->uPause(p, 150);
+		}
+		else
```
haraldg: No need for empty else statements. Just drop it.

server/drivers/hd44780.c
```diff
@@ -157,6 +157,7 @@ HD44780_init(Driver *drvthis)
 	p->line_address 	= drvthis->config_get_int(drvthis->name, "lineaddress", 0, LADDR);
 	p->have_keypad		= drvthis->config_get_bool(drvthis->name, "keypad", 0, 0);
 	p->have_backlight	= drvthis->config_get_bool(drvthis->name, "backlight", 0, 0);
+	p->isVFDDisplay		= drvthis->config_get_bool(drvthis->name, "VFDDisplay", 0, 0);
 	p->have_output		= drvthis->config_get_bool(drvthis->name, "outputport", 0, 0);
```
haraldg: Is the register set really generic to all VFD displays or just an extension of the PT6314  only. If it is the latter, please use a less generic name for both the configuration option and the variable. Ie something containing "pt6314".

haraldg (changes requested):
I added inline comments to one connection type, but of course they apply to all of them.

I notice, in some connection types you call uPause, while in othes you don't. Is there a reason for that?

I would still prefer to implement this in the common part of the driver instead of each connection type. But if it is difficult to do for some reason, that I'm ok with changing the connection types. However testing this with all connection types won't happen ...

You changed access permissions for a number of files and added files, that really don't belong into git. We will need to clean this up before merging as well.

trex2000:
Will do the requested changes. 
To answer your questions:
- I saw th uPause being called in the respective module also in other cases , therefore I've added also here. In some other sub-modules, it wasn't, so i've skipped it there.
- I will change 'isVFDDisplay' to 'isPT6314VFDDisplay'
- i have an automotive background and here we're required to use the smallest visibility scope of variables and empty else statements. Therefore I always add the empty elses, it's required by Misra guidlines. Regarding adding the variable to private_data.. Hmm, if you want it that way, I can add it, however I don't really see the point. That's just a 'temporary' variable that needs to be visible only in that module, however as it's necessary to be used in 2 separate functions, I couldn't define it locally in one of the functions. I did not want also to modify the parameter list of the functions just for this. This was the reason why I've declared it at file-level.

I could use some help with cross compiling. I would like to build it for the ARM architecture, to be able to test it on libreelec on an odroid c2, which I currently have and has such a display connected.

haraldg:
> I saw th uPause being called in the respective module also in other cases , therefore I've added also here. In some other sub-modules, it wasn't, so i've skipped it there.

Well different authors have worked on the code and the various connection types over time. It's a messy code base, but since all connection types are meant to driver the same controller, the timing should be the same I think. If there is no known reason why a pause should be added, I recommend not adding it by default.

> i have an automotive background and here we're required to use the smallest visibility scope of variables

This is a good habit, but the driver API doesn't allow it. IE the API requires driver functions to be reentrant and having variables with file scope isn't. The smallest possible "scope" is as private data.

> and empty else statements

lcdproc coding style doesn't allow this - I rarely agree with our coding style, but this time I do.

please drop them

> I could use some help with cross compiling. I would like to build it for the ARM architecture, to be able to test it on libreelec on an odroid c2, which I currently have and has such a display connected.

Hm, which help do you need?

I am cross compiling for older ARM systems, but for recent systems I'm running debian testing (armhf) and do all compiling natively.

trex2000:
I've done the corrections. Could you review it ?

haraldg:
I think this is looking quite good. I'd still prefer if we could add this to the main driver instead of all the connection types, but that's not a problem.

I think the old_state variables really should be initialized with some clearly invalid value. AFAICS 0 is actually valid?

Also we need to document the new configuration setting in the user guide and maybe add it (commented out) in the example configuration.

Since you seem somewhat uncomfortable with git, I can do part of the cleanup of the PR for you, but please drop all the unnecessary whitespace changes.

trex2000:
> "I think this is looking quite good. I'd still prefer if we could add this to the main driver instead of all the connection types, but that's not a problem."

I've analyzed the code but it seems that not all driver types communicate directly with the HD44780's registers. Therefore I had to add it separately to each 'connection type'.

> I think the old_state variables really should be initialized with some clearly invalid value. AFAICS 0 is actually valid?
It's not a problem with the old state being 0. Normally after startup, in the init funtion the backlight  bits in the FUNCSET register will be set to 0.
The p->backlight attribute will be 0. 
In case the backlight is requested, this will become 1, therefore there will be a rising edge of the old state. 
This is the moment when we have to update the FUNCSET register with the new brightness level. I've added this to prevent cyclic updates of the FUNCSET register. I don't think this would be a problem either, but if it's not necessary, then why doing it ?

> Also we need to document the new configuration setting in the user guide and maybe add it (commented out) in the example configuration.
Sure, shall I update also these files ?


> Since  you seem somewhat uncomfortable with git, I can do part of the cleanup of the PR for you, but please drop all the unnecessary whitespace changes.
I've tried to check the differences between the version before my updates and after. But it seems git is quite 'commit' based change management software, and not file based. I couldn't find how to do it in one step, to list all the files changed with my 3-4 commits.
Is there a way ?

haraldg:
On 15.09.2017 13:30, trex2000 wrote:
>> "I think this is looking quite good. I'd still prefer if we could 
>> add this to the main driver instead of all the connection types, but 
>> that's not a problem."
>
> I've analyzed the code but it seems that not all driver types
> communicate directly with the HD44780's registers. Therefore I had to
> add it separately to each 'connection type'.

Can you elaborate on this? Which drivers would that be? How do they 
work
instead?

I'm not aware which connection types that would be, but then I don't 
have
a HD44780 myself, so I don't know much about this driver.

>> I think the old_state variables really should be initialized with 
>> some clearly invalid value. AFAICS 0 is actually valid?
> It's not a problem with the old state being 0. Normally after
> startup, in the init funtion the backlight  bits in the FUNCSET
> register will be set to 0.
> The p->backlight attribute will be 0.
> In case the backlight is requested, this will become 1, therefore
> there will be a rising edge of the old state.
> This is the moment when we have to update the FUNCSET register with
> the new brightness level. I've added this to prevent cyclic updates 
> of
> the FUNCSET register. I don't think this would be a problem either,
> but if it's not necessary, then why doing it ?

It's just more obviously correct.

(BTW: I think the backlight function of the main driver already
suppresses redundant calls, so probably keeping track of backlight
state won't be necessary anyway.)

>> Also we need to document the new configuration setting in the user 
>> guide and maybe add it (commented out) in the example configuration.
> Sure, shall I update also these files ?

Yes.

>> Since  you seem somewhat uncomfortable with git, I can do part of 
>> the cleanup of the PR for you, but please drop all the unnecessary 
>> whitespace changes.
> I've tried to check the differences between the version before my
> updates and after. But it seems git is quite 'commit' based change
> management software, and not file based. I couldn't find how to do it
> in one step, to list all the files changed with my 3-4 commits.
> Is there a way ?

Actually there are many ways:
Just viewing the changes: 'git diff oldcommit newcommit'
Actually merging multiple commits into one: 'git merge --squash ...'

HTH,
Harald

-- 
If you want to support my work:
see http://friends.ccbib.org/harald/supporting/
or donate via peercoin to P98LRdhit3gZbHDBe7ta5jtXrMJUms4p7w
or CLAM xASPBtezLNqj4cUe8MT5nZjthRSEjrRQXN

haraldg:
I guess this PR is obsolete by #95 now, so closing ...

PR #90, opened 2017-09-08
Add missing ioctl header

baruchsiach:
The spidev.h header uses macros from linux/ioctl.h. Add this header
explicitly since some libc, like musl, do not include it implicitly.
This fixes the following build failure:

In file included from .../sysroot/usr/include/sys/ioctl.h:7:0,
                 from hd44780-spi.c:31:
hd44780-spi.c: In function ‘spi_transfer’:
hd44780-spi.c:89:24: error: ‘_IOC_SIZEBITS’ undeclared (first use in this function)
  status = ioctl(p->fd, SPI_IOC_MESSAGE(1), &xfer);
                        ^

### DIFF
--- a/server/drivers/hd44780-pifacecad.c
+++ b/server/drivers/hd44780-pifacecad.c
@@ -54,6 +54,7 @@
 #include <unistd.h>
 #include <fcntl.h>
 #include <sys/ioctl.h>
+#include <linux/ioctl.h>
 #include <stdint.h>
 #include <linux/spi/spidev.h>
 
--- a/server/drivers/hd44780-spi.c
+++ b/server/drivers/hd44780-spi.c
@@ -29,6 +29,7 @@
 #include <unistd.h>
 #include <fcntl.h>
 #include <sys/ioctl.h>
+#include <linux/ioctl.h>
 #include <stdint.h>
 #include <linux/spi/spidev.h>
 


haraldg:
Are you sure this isn't a bug either in musl or the
linux headers?

I think including one header should be safe without
including an other?

baruchsiach:
The kernel included `tools/spi/spidev_test.c` tool includes `linux/ioctl.h` in addition to `linux/spi/spidev.h`. Both headers are Linux specific, so this has nothing to do with musl. You might argue that `spidev.h` should include the Linux specific `ioctl.h`. But event if that gets fixed in the kernel, you would still like to keep compatibility with older kernel headers.

haraldg:
Yes, I think we will merge this for backwards compatibility. However I want to make sure that if there is a bug elsewhere, it get's properly reported and fixed therer too.

baruchsiach:
Posted a patch for the kernel header.

https://marc.info/?l=linux-spi&m=150506459822531&w=2

haraldg:
Looks like nobody responded to the patch to the kernel so far. I have picked up the patch to lcdproc with the workaround, because I don't see the point in waiting any longer. However leaving this PR open as a remainder, that this still isn't fixed properly.

baruchsiach:
The kernel patch is currently queued in the the SPI maintainer's git tree in a fixes for v4.14 branch. So this should be fixed in the final v4.14. Mark even marked it for stable kernels.

https://git.kernel.org/pub/scm/linux/kernel/git/broonie/spi.git/commit/?h=fix/uapi&id=a2b4a79b88b24c49d98d45a06a014ffd22ada1a4

haraldg:
Nice. I didn't see any confirmation in the linux-spi ML archive. Closing this now.

Issue #79, opened 2017-04-19
Driver Code Consolidation and a "Universal" Driver


fmertz:
In the spirit of moving this project forward and keeping changes incremental, how about combining all (or most) existing drivers into just one "universal" driver (yes, one more driver, for now).

Let's get this out of the way first: [https://xkcd.com/927/](url)

This new driver would implement a matrix, by de-coupling the "language" from the "hardware interface", meaning this driver could be told what command set to use on the one hand (the few drivers I have seen seem to have similar command sets), and then separately what particular hardware interface to issue those commands over. This new driver could have all the interface code compiled-in or use some new ABI to dynamically load one interface at a time to minimize memory footprint on embedded platforms. 

Remember, there is no computer problem that can't be simplified by adding a level of indirection!

If we were to implement something like this, I believe the majority of the needed code would be already there in the existing implementation of each driver. There would be a need for a (massive) refactoring. The HD driver(s) seem like a great place to start.

Also, it would not be disruptive as it would not require changes to LCDd, or ABI (unless we wanted to). It could also be a phased implementation, with more and more LCDs supported over time. This would allow us to retire the corresponding discrete drivers over time, too.

The main idea is that it could fill in the matrix of combinations "for free" and avoid silly situations where we have a driver with the right commands here, and another driver for the hardware interface there, but no implementation combining the two.

Also, we could add "high level" configurations entries that would describe popular combinations of LCDs and interface for pre-made hardware, like commercial firewalls, e.g. an entry for "Watchguard/Lanner" would know to engage the SDEC "language" and the parallel port "interface", making it EASY for end users to get their hardware going, and be the next best thing to an automatic detection. The idea here is to make it easy for individual distributions to create glue code to integrate LCDproc. An example is pfSense, which needs PHP code to integrate packages into the GUI of the distribution. A complicated set of drivers and parameters makes it hard for users to get things to work. If there were fewer options, and pre-configured known-good combinations, this glue-code would be easier to write and maintain, and eventually serve the users better.

Thanks for any thought.

haraldg:
I think one universal driver probably is too much -- modularity is a good thing after all, but our current amount of code duplication hints that modularity is done in a poor way.

Your proposal of a language/hardware interface matrix is pretty close to what we have been discussing in #49. I think we have done there about 50% of what's necessary to propose an API suitable for implementing such a decoupling. Maybe you can start from there?

[...]

Issue #49, opened 2017-01-18
Feature Request - Parallel port - RPi GPIO compatibility wrapper?


JamesGKent:
I'm looking to use a graphical LCD (with a T6963 driver) on a raspberry pi, however it seems that the only driver for this is written to use the parallel port. i know that the HD44780 type displays have had the driver adapted to use the GPIO pins of the pi with the correct connection type set in the driver config.
Given that a lot of the drivers are written to use either USB or a parallel port i'm wondering if it would be possible to write a compatibility wrapper to translate parallel port calls to GPIO in a common way to prevent having to rewrite each driver to use the GPIO?

I would attempt to adapt the driver i need but i'm a novice with C/C++ and having had a look at the driver files i'm not even sure where to begin...

haraldg:
Having a more modulare architecture of lcdproc and a hardware abstraction layer for the different kinds of ports would be possible and actually a very good thing. However, I don't think anybody is working on that.

If you want to work on a new connection type for the T6963 driver have a look at t6963_low.c which is just above 200 lines of code. Also use hd44780-gpio.c as a reference how to use gpios.

Since this is a graphical lcd it would also be reasonable to write a driver for a linux framebuffer or fbtft device. Of course submitting a kernel driver might be a daunting task for somebody new to C and just writing a replacement or improvement for t6963_low.c will be a lot less frustrating.

JamesGKent:
further to your comment about the unmaintained patches, and your comment about using ppdev as a base for a hardware abstraction wrapper.
I was thinking that it should be possible to write a generic wrapper that emulates a parallel port (for unmodified drivers) and maps the pins present in a parallel port to the linux sysfs gpio interface, that way the wrapper would work for more than just the RPi GPIO and be more useful.
however I can't see a clean way to do this without defining a new section in the config such as:
```
[output]
# valid values: parallel, gpio
Type=gpio

# valid only when Type=parallel
Address=0x3F

# valid only when Type=gpio
# used to map parallel port pins to GPIO pins
pin_P_Strobe=1
pin_P_D0=2
pin_P_D1=3
pin_P_D2=4
pin_P_D3=5
pin_P_D4=6
pin_P_D5=7
pin_P_D6=8
pin_P_D7=9
pin_P_Ack=10
pin_P_Busy=11
pin_P_PaperOut=12
pin_P_Select=13
pin_P_Linefeed=14
pin_P_Error=15
pin_P_Reset=16
pin_P_SelectPrinter=17
[...]

PR #94, opened 2017-09-25
Add support for HD44780-compatible WINSTAR WEH00xxyyA OLED Displays

mskalski:
Added support for WINSTAR OLED displays - almost compatible to
HD44780 LCDs. These displays require slightly different initialization sequence,
especially required on display reinitialization without powering it off first.

Also added support for handling backlight using built-in command
for brightness setting.

Tested on Raspberry PI and WINSTAR WEH001604ALPP5N00001 OLED display
connected through I2C PCF8574 connection (4-bits obviously) and
WINSTAR WEH002004ALPP5N00001 OLED. Should work on any connection.

### DIFF
--- a/LCDd.conf
+++ b/LCDd.conf
@@ -598,6 +598,28 @@ Size=20x4
 # than one display!
 #ExtendedMode=yes
 
+# Some OLED displays (particularly WINSTAR WEH001602A, WEH002002A, and others)
+# require slightly different initialization sequence, which is particularly important
+# on reinitializations without powering off display. If you have this kind 
+# of display and have garbage on it after restarting LCDd, turn this option to TRUE.
+# By default it is FALSE.
+#WinstarMode=yes
+
+# WINSTAR OLED WEH001602A and compatible have two levels of brightness 
+# (full and dimmed), which can be set using special LCD command, not with
+# dedicated backlight pin. This option if enabled maps backlight setting 
+# to these two levels of brightness, depending on values of options (respectively)
+# Brightness and OffBrightness, any value below 500 causes dimmed brightness;
+# >= 500 causes full brightness.
+#
+# If this option is OFF, brightness setting is not managed with backlight, but it is
+# set to constant value at time of initialization, depending on configured Brightness
+# setting.
+#
+# By default it is set, as it appears to be common usage. Works only when WinstarMode
+# is turned ON, otherwise it is ignored.
+WinstarMapBl=no
+
 # In extended mode, on some controllers like the ST7036 (in 3 line mode)
 # the next line in DDRAM won't start 0x20 higher. [default: 0x20]
 #LineAddress=0x10
--- a/docs/lcdproc-user/drivers/hd44780.docbook
+++ b/docs/lcdproc-user/drivers/hd44780.docbook
@@ -3772,6 +3772,57 @@ This can be done by specifying <option>--enable-drivers=all</option> or by inclu
 
 <varlistentry>
   <term>
+    <property>WinstarMode</property> = &parameters.yesnodef;
+  </term>
+  <listitem><para>
+      Some OLED displays (particularly WINSTAR WEH001602A, WEH002002A, and others)
+      require slightly different initialization sequence, which is particularly important
+      on reinitializations without powering off display. If you have this kind 
+      of display and have garbage on it after restarting LCDd, turn this option to TRUE.
+    
+      Default: <literal>no</literal>.
+  </para></listitem>
+</varlistentry>
+
+<varlistentry>
+  <term>
+    <property>WinstarMapBl</property> = &parameters.yesnodef;
+  </term>
+  <listitem>
+    <para>
+      This option works only when <literal>WinstarMode</literal> is turned ON, otherwise it is ignored.
+    </para>
+    <para>
+      WINSTAR OLED WEH001602A and compatible have two levels of brightness 
+      (dimmed and full), which can be set using special LCD command, not with dedicated
+      backlight pin. This option if enabled maps backlight setting to these two levels of brightness,
+      depending on values of options <literal>Brightness</literal> and <literal>OffBrightness</literal>:
+      <itemizedlist>
+        <listitem><para>
+          <literal>&lt;500</literal> causes dimmed brightness
+        </para></listitem>
+        <listitem><para>
+          <literal>&gt;=500</literal> causes full brightness.
+        </para></listitem>
+      </itemizedlist>
+    </para>
+    <para>
+      If this option is OFF, brightness setting is not managed with backlight, but it is
+      set to constant value at time of initialization, depending on configured setting
+      of <literal>Brightness</literal>.
+  </para>    
+    <para>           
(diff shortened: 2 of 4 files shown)


haraldg:
Looks almost perfect to me. The only thing missing is updating the LCDd.conf with the new (commented out) options you are introducing.

I'd have prefered the initialization part and the backlight part to be different commits, but never mind that now.

@trex2000 this PR seems to do something very similar to what might work for your PR #89 too. AFAICS only the register values are differnt. Maybe you can have a look and tell us, if this code can be generalized to cover your case as well? I'm thinking along the lines of renaming WinstarMapBl to something like InternalBl and introduce a new configuration setting for the actual register value.

mskalski:
Thanks for review

I thought about separating commits, but actually initialization depends on value of WinstarMapBl to set correct brightness.

Yes, I forgot about example in LCDd.conf, I'll fix it, thanks for pointing this out.

As you probably noted I slightly changed order of sending commands, which are sent to HDI44780 controller, so they may influence other compatible displays. I tested my change also on one connected through i2c backpack (I had no other hardware to test with) and it works well, but maybe someone else could test it also?

I thought a little bit more about my PR, and I think it can be done better:

1. Always on initialization turn off display (not only for Winstar OLED), to hide possibly displaying garbage. It should be change compatible to other displays.
2. Make WindstarMapBl independent of Backlight setting, to allow simultaneously handling internal brightness and external BL if someone needs it, but I feel it is a little overkill.

Any comments are welcome.😀

Regarding #89:
What do you mean setting register value?  In config file? If yes, I think it could be good idea for advanced users, but most (I think) users just know type of display and are not aware of commands required to handle internals. Feature I added was quite hidden, just found one note on some forum, I can't even remember where, during searching resolution for problems with reinitialization of WindstarMapBl OLED; In manual it's not documented, just mentioned.

So I think there should be options to turn on specific features (documented ie. in lcdproc manual) and they shall just work. If anyone would be interested in, he can look to source code how it is achieved.

haraldg:
re testing: I don't have a HD44780 so I can't help you there. You can ask on the ML or directly request feedback from previous commiters to the driver in this PR. I will wait with merging the PR if you think more testing is necessary.

re 1: Sounds reasonable

re 2: I think your code is fine as it is. But you know the display best and know what's useful and what is not.

re #89: Yes, I mean in config file. I'm not thinking about advanced users, but about reducing code duplication. For now we would have the register values commented out in the configuration file and the user would uncomment the section describing his display. In the long run I hope to extend the configuration language like outlined in https://github.com/lcdproc/lcdproc/issues/49#issuecomment-280653594 to make this even more comfortable.

mskalski:
re #89: I think it is similar to this PR, but slightly different. _PT6314 VFD_ has four levels of brightness, _WINSTAR OLED_ has two. To do things right, these levels shall be mapped to configured `Brightness` and `OffBrightness` and I don't see a way to do it straight from configuration file. Especially it can  possibly be called at different stages, not only on backlight setting but also during initialization.

I think it is safer and clearer to add code to driver and one more option which enables this behavior (as I did with `WinstarMapBl`). My change was easier as it is caused by a change in (possibly incompatible) initialization sequence and requirements, so for this introduced `WinstarMode` option, so when former is enabled it means someone has this kind of screen and first is only as a side effect.  :smile:

Change for VFD can be by default disabled, as it is not handled by other display types and sometimes can be .

Or we can introduce something like `Subtype` or `SubModel` option, something similar to ideas in #49 or #79, which handle _Command Set differencies_ for HD44780 display(s). There are not so many now yet though: `ExtendedMode`, `WinstarMode` for now and new _VFD PT6314_.

On the other hand I think your proposition of introducing options for defining command(s) to on and off backlight is worth of implementing as last resort solution for future(maybe even unknown yet) displays.

trex2000:
Thanks for joining these 2 commits. 
So, for the PT6314 driver, there are 2 bits in additionin FUNCSET register, that handle the VFD brightness:
00 - 100%
01 - 75%
02 - 50%
03 - 25%
So there are 4 different brightness levels.
In my commits, I've mapped these to the backlight setting rather than brightness, as usually (also XBMC plugin does this) the backlight (on or off) is set by client during runtime. This way, when there is an 'event' the backlight was turned on, and after a timeut, the backlight was turned off. For the VFD, this meant that in case of an 'event', the XBMC plugin requested 100% 
brightness on the VFD and after a while the 'backlight' was turned off, so the VFD's brigthness was set back to minimum value, 25%. This give a similar behavior like any commercial AV receiver has.
Now, the problem was that the FUCSET register was set once during init time, as it contained no information that was supposed to be set during runtime. But with brightness this changed, therefore I had to store the value of the FUNCSET register that was set at init, and use it also during runtime when the backlight request was changed. 
This is the modification that I've done.
Indeed, yours, for the OLED looks very similar, so I would also propose to somehow merge the implementation.

http://www.artronic.eu/pl/p/VFD-AC-1602H-BG-5V/937
This is the display I'm using and here's the datasheet:
https://www.newhavendisplay.com/app_notes/PT6314.pdf

trex2000:
Yes, also to me, instead of 100% brightness, maybe it's safer to have it initialized with 25% only, therefore value 11. I wouldn't turn if off in this case, as by default nobody would set the parameter correctly, therefore the display won't be initialized. Also the 00 means 100% , therefore maximum brightness, and I think also the HW manufactureres thought it this way. In case there is no 'differentiated' support (so only generic instruction set for the HD44780 is available) the display would still work, but without brightness control.

mskalski:
I was thinking about above comments and these are my propositions:

Introduce `Model` option, as in other display types. It should be used to select command differences incompatible between various hd44780 clones. Here now would be `standard`, `winstar_oled`, `pt6314_vfd` or maybe more generic names. For now `pt6314_vfd` is only defined for handling internal backlight. Value `winstar_oled` causes slightly different initialization sequence and allows to choose internal backlight commands.

Modify existing `Backlight` option from Boolean to enumeration: `none`, `external`, `internal`,`internalCmds`and with currently existing Boolean values `yes`/`no` mapped to `external` and `none`.

* `internal` means that backlight is handled using internal commands according to selected model. Specific commands would be coded according to `Brightness` and `OffBrightness` options.

* `internalCmds` would mean that commands for turning on and off backlight are given with options `BacklightOnCmd` and `BacklightOffCmd`, which would be treated as catch up (last resort) for other types of displays which have similar features.

* `external` means same as for now `yes` - handling backlight using external pin.

I would rather not left only these `internalCmds` options. First they may interfere with standard initialization (for example for VFD one need to define interface bit lenght 4/8 bits additionally to specifying brightness). Second it's impossible to handle brightness set through driver interface this way - functions `get_brightness()` and `set_brightness()`.

It is tempting to allow combinations of some above options, to allow hypothetically possible situation of setting backlight  with both means (ie. bright display and extra led for glowing buttons) :smile:

I'll make new pull request with these proposed changes, with commits for enabling model selection and then backlight handling.

haraldg:
I have to admit that since there is no way to change the brightness from the client API, I completely forgot that we brightness control in the driver API. With that in mind I guess your approach makes sense.

PR #95, opened 2017-09-30
Add support for HD44780-compatible WINSTAR WEH00xxyyA OLED Displays - try 2

mskalski:
Added support for WINSTAR OLED displays - almost compatible to
HD44780 LCDs. These displays require slightly different initialization sequence,
especially required on display reinitialization without powering it off first.

This was achieved through additional Model option, incorporating to it 
existing 'ExtendedMode' for KS0073 display which require also special
handling. Added also PT6314 VFD model to available models.

Tested on Raspberry PI and WINSTAR WEH001604ALPP5N00001 OLED display
connected through I2C PCF8574 connection (4-bits obviously) and
WINSTAR WEH002004ALPP5N00001 OLED. Should work on any connection.

It is first part - next would be adding backlight handling using internal brightness commands for WINSTAR OLED and PT6314 VFD.

### DIFF
--- a/server/drivers/hd44780-low.h
+++ b/server/drivers/hd44780-low.h
@@ -126,6 +126,9 @@
 #define DEFAULT_OFFBRIGHTNESS	300
 /**@}*/
 
+/** Maximum value of brightness */
+#define MAX_BRIGHTNESS		1000
+
 /** \name Maximum sizes of the keypad
  *@{*/
 /* DO NOT CHANGE THESE VALUES, unless you change the functions too! */
@@ -277,6 +280,15 @@ typedef struct hd44780_private_data {
 	          		     set to HD44780_MODEL_EXTENDED */
 	int line_address;	/**< address of the next line in extended mode  */
 	int backlight_type;	/**< way of handling backlight. */
+	int backlight_cmd_on;	/**< internal command(s) for enabling backlight */
+	int backlight_cmd_off;	/**< internal command(s) for disabling backlight */
+
+	/** Value saved during display initialization in common_init(),
+	 *  given for FUNC_SET command, to use for later.
+	 *
+	 *  NOTE: if common_init() from specific connection type is not called,
+	 *  update it for correct value. For now used only for PT6314_VFD model */
+	int func_set_mode;
 
 	int delayMult;		/**< Delay multiplier for slow displays */
 	char delayBus;		/**< Delay if data is sent too fast over LPT port */
@@ -494,6 +506,7 @@ static void set_have_backlight_pin(PrivateData *p, int on) {
 #define WINST_PWRON	0x04	/**< Internal power on (high brightness)*/
 #define WINST_PWROFF	0x00	/**< Internal power off (low brightness)*/
 
+
 /** Function set (RE=0) */
 #define FUNCSET		0x20
 #define IF_8BIT		0x10
@@ -504,6 +517,18 @@ static void set_have_backlight_pin(PrivateData *p, int on) {
 #define SMALLCHAR	0x00	/**< 5x8 characters */
 #define EXTREG		0x04	/**< Select ext. registers (Yes, the same bits) */
 #define SEGBLINK	0x02	/**< CGRAM/SEGRAM blink, only if RE=1 */
+
+/** Extra definitions for setting PT6314_VFD brihtness
+ *  yes - same bits used as extended registers */
+#define PT6314_BRIGHT_100	0x00
+#define PT6314_BRIGHT_75	0x01
+#define PT6314_BRIGHT_50	0x02
+#define PT6314_BRIGHT_25	0x03
+
+/* mask for brightness */
+#define PT6314_BRIGHT_MASK	0x03
+
+
 
 /** Set CGRAM address (RE=0) */
 #define SETCHAR		0x40
--- a/server/drivers/hd44780-low.h
+++ b/server/drivers/hd44780-low.h
@@ -97,6 +97,28 @@
 
 /** @} */
 
+/** \name Types of backlight handling
+ * @{
+ * Shall be treated as bitmask
+ */
+
+
+/** No backlight handling at all */
+#define BACKLIGHT_NONE              0
+
+/** Backlight handled hrough external pin. */
+#define BACKLIGHT_EXTERNAL_PIN 0x0001
+
+/** Backlight handled through internal commands built in driver,
+ *  usually through special commands for model chosen with Model option*/
+#define BACKLIGHT_INTERNAL     0x0002
+
+/** Backlight handled through internal commands defined in configuration file */
+#define BACKLIGHT_CONFIG_CMDS  0x0004
+
+/** @} */
+
+
 /** \name Symbolic default values
  *@{*/
 #define DEFAULT_CONTRAST	800
@@ -246,14 +268,15 @@ typedef struct hd44780_private_data {
 	/** \name Display features
 	 *@{*/
 	char have_keypad;
-	char have_backlight;
 	char have_output;
+	/* have_backlight moved to function have_backlight_pin() below */
 	/**@}*/
 
 	int model;		/**< model selected in configuration.
 	          		     For extended mode on some weird controllers
 	          		     set to HD44780_MODEL_EXTENDED */
 	int line_address;	/**< address of the next line in extended mode  */
+	int backlight_type;	/**< way of handling backlight. */
 
 	int delayMult;		/**< Delay multiplier for slow displays */
 	char delayBus;		/**< Delay if data is sent too fast over LPT port */
@@ -394,6 +417,19 @@ has_extended_mode(PrivateData *p) {
 	return p->model == HD44780_MODEL_EXTENDED;
 }
 
+/* returns if display is configured to use external backlight pin */
+static char have_backlight_pin(PrivateData *p) {
+
+	return (p->backlight_type & BACKLIGHT_EXTERNAL_PIN);
+}
+
+/* sets configration value for using external pin for backlight */
+static void set_have_backlight_pin(PrivateData *p, int on) {
+	if (on)
+		p->backlight_type |= BACKLIGHT_EXTERNAL_PIN;
+	else
+		p->backlight_type &= ~BACKLIGHT_EXTERNAL_PIN;
+}
 
 
 /* commands for senddata */
--- a/server/drivers/hd44780-low.h
+++ b/server/drivers/hd44780-low.h
@@ -388,6 +388,13 @@ typedef struct hwDependentFns {
 /* Prototypes */
 void common_init(PrivateData *p, unsigned char if_bit);
 
+/* returns if display needs/is using 'Extended mode'- nonzero if yes */
+static short
+has_extended_mode(PrivateData *p) {
+	return p->model == HD44780_MODEL_EXTENDED;
+}
+
+
 
 /* commands for senddata */
 #define RS_DATA		0x00
--- a/server/drivers/hd44780-low.h
+++ b/server/drivers/hd44780-low.h
@@ -129,6 +129,10 @@
 /** Maximum value of brightness */
 #define MAX_BRIGHTNESS		1000
 
+/** Maximum value of contrast */
+#define MAX_CONTRAST		1000
+
+
 /** \name Maximum sizes of the keypad
  *@{*/
 /* DO NOT CHANGE THESE VALUES, unless you change the functions too! */
--- a/server/drivers/hd44780-low.h
+++ b/server/drivers/hd44780-low.h
@@ -75,6 +75,28 @@
 #define IF_TYPE_SPI		6
 /**@}*/
 
+/** \name Symbolic name for specific models
+ * @{
+ * Used here to handle various conflicting differences in command sets.
+ * Some optional features are nottied to model even if it appears only in that model
+ * unless not are not conflicting, as default settings are compatible across all devices.
+ * (for example FontBank) */
+
+/** Standard / default model. Should work with most of HD44780-compatible displays */
+#define HD44780_MODEL_DEFAULT		0
+/** Extended model - 4 lines activated with special bits and extra commands (EXTREG) */
+#define HD44780_MODEL_EXTENDED		1
+/** WINSTAR WEH00xxyyA (WEH001604A, WEH002004A, ...) and others, requires slightly different
+ * initialization sequence and have extra commands for handling brightness */
+#define HD44780_MODEL_WINSTAR_OLED	2
+/** PTC PT6314 VFD Displays - have extra command for setting four-level brightness.
+ *  Otherwise is compatible to other HD44780s */
+#define HD44780_MODEL_PT6314_VFD	3
+
+/* Possibly extended further...*/
+
+/** @} */
+
 /** \name Symbolic default values
  *@{*/
 #define DEFAULT_CONTRAST	800
@@ -162,7 +184,7 @@ typedef struct hd44780_private_data {
 
 #ifdef HAVE_I2C
 	/* i2c based connection types */
-	
+
 	int i2c_backlight_invert;
 	int i2c_line_RS;
 	int i2c_line_RW;
@@ -228,7 +250,9 @@ typedef struct hd44780_private_data {
 	char have_output;
 	/**@}*/
 
-	char ext_mode;		/**< use extended mode on some weird controllers */
+	int model;		/**< model selected in configuration.
+	          		     For extended mode on some weird controllers
+	          		     set to HD44780_MODEL_EXTENDED */
 	int line_address;	/**< address of the next line in extended mode  */
 
 	int delayMult;		/**< Delay multiplier for slow displays */
@@ -419,6 +443,13 @@ void common_init(PrivateData *p, unsigned char if_bit);
 
 /** Shift or scroll enable (RE=1) */
 #define HSCROLLEN	0x10
+
+/** Extra definitions on Winstar OLED displays - set last 2 bits (=0x03) to activate */
+#define WINST_MODESET	0x13	/**< required bits for command */
+#define WINST_TEXTMODE	0x00	/**< Activate text mode */
+#define WINST_GRAPHMODE	0x08	/**< Activate graphic mode */
+#define WINST_PWRON	0x04	/**< Internal power on (high brightness)*/
+#define WINST_PWROFF	0x00	/**< Internal power off (low brightness)*/
 
 /** Function set (RE=0) */
 #define FUNCSET		0x20
--- a/server/drivers/hd44780.c
+++ b/server/drivers/hd44780.c
@@ -304,6 +304,13 @@ HD44780_init(Driver *drvthis)
 		report(RPT_ERR, "%s: unknown Backlight type: %s", drvthis->name, s);
 		return -1;
 	}
+	p->backlight_cmd_on	= drvthis->config_get_int(drvthis->name, "backlightcmdon", 0, 0);
+	p->backlight_cmd_off	= drvthis->config_get_int(drvthis->name, "backlightcmdoff", 0, 0);
+	if ((p->backlight_type & BACKLIGHT_CONFIG_CMDS) && (!p->backlight_cmd_on || !p->backlight_cmd_off)) {
+		report(RPT_ERR, "%s: No commands for enabling or disabling backlight specified for backlight type internalCmds", drvthis->name);
+		return -1;
+	}
+
 	p->have_output		= drvthis->config_get_bool(drvthis->name, "outputport", 0, 0);
 	p->delayMult 		= drvthis->config_get_int(drvthis->name, "delaymult", 0, 1);
 	p->delayBus 		= drvthis->config_get_bool(drvthis->name, "delaybus", 0, 1);
@@ -616,16 +623,45 @@ HD44780_init(Driver *drvthis)
 void
 common_init(PrivateData *p, unsigned char if_bit)
 {
+	/* If currently backlightstate is not yet initialized, it is -1, which is != 0,
+	 * so set brightness based on Brightness setting. This will work also if backlight handling
+	 * is not used */
+	int init_brightness = (p->backlightstate) ? p->brightness : p->offbrightness;
+
+	unsigned char cmd_funcset =  FUNCSET | if_bit | TWOLINE | SMALLCHAR;
 	if (has_extended_mode(p)) {
 		/* Set up extended mode */
-		p->hd44780_functions->senddata(p, 0, RS_INSTR, FUNCSET | if_bit | TWOLINE | SMALLCHAR | EXTREG);
+		p->hd44780_functions->senddata(p, 0, RS_INSTR, cmd_funcset | EXTREG);
 		p->hd44780_functions->uPause(p, 40);
 		p->hd44780_functions->senddata(p, 0, RS_INSTR, EXTMODESET | FOURLINE);
 		p->hd44780_functions->uPause(p, 40);
 	}
 
-	/* set up standard mode. by default font_bank is zero (and is ignored on most of displays) */
-	p->hd44780_functions->senddata(p, 0, RS_INSTR, FUNCSET | if_bit | TWOLINE | SMALLCHAR | p->font_bank);
+	if (p->model == HD44780_MODEL_PT6314_VFD) {
+		cmd_funcset &= ~PT6314_BRIGHT_MASK;
+		p->func_set_mode = cmd_funcset;
+
+		if (init_brightness >= 3 * (MAX_BRIGHTNESS / 4))
+			cmd_funcset |= PT6314_BRIGHT_100; /* = 0x00 */
+		else if (init_brightness >= MAX_BRIGHTNESS / 2)
+			cmd_funcset |= PT6314_BRIGHT_75; /* = 0x01 */
+		else if (init_brightness > MAX_BRIGHTNESS / 4)
+			cmd_funcset |= PT6314_BRIGHT_50; /* = 0x02 */
+		else
+			cmd_funcset |= PT6314_BRIGHT_25; /* = 0x03 */
+	}
+	else {
+		/* by default font_bank is zero
+		 * this is ignored on most of displays, except of PT6314 VFD,
+		 * which this value means completely different, but is handled above */
+		cmd_funcset |=  p->font_bank;
+	}
+
+	/* save used cmd for FUNCSET */
+	p->func_set_mode = cmd_funcset;
+
+	/* set up standard mode.  */
+	p->hd44780_functions->senddata(p, 0, RS_INSTR, cmd_funcset);
 	p->hd44780_functions->uPause(p, 40);
 
 	/* Turn off display, as manipulatimg below can cause some garbage on screen */
@@ -640,7 +676,7 @@ common_init(PrivateData *p, unsigned char if_bit)
 		/* For WINSTAR OLED displays need to set TEXT mode and additionally level of brigtness.
 		 * It is particularly important on reinitialization without powering off it first */
 		unsigned char pwr = WINST_PWROFF;
-		if (p->brightness >= 500) {
+		if (init_brightness >= MAX_BRIGHTNESS / 2) {
 			pwr = WINST_PWRON;
 		}
 		p->hd44780_functions->senddata(p, 0, RS_INSTR, WINST_MODESET | WINST_TEXTMODE | pwr);
@@ -1032,7 +1068,7 @@ HD44780_set_brightness(Driver *drvthis, int state, int promille)
 	PrivateData *p = drvthis->private_data;
 
 	/* Check it */
-	if (promille < 0 || promille > 1000)
+	if (promille < 0 || promille > MAX_BRIGHTNESS)
 		return;
 
 	/* store the software value since there is not get */
@@ -1050,6 +1086,60 @@ HD44780_set_brightness(Driver *drvthis, int state, int promille)
 	p->backlightstate = -1;
 }
 
+/** Sets internal backlight state using internal commands (based on configured model) */
+static void
+hd44780_set_backlight_internal(PrivateData *p, int state)
+{
+	unsigned char cmd = 0;
+	int brightness = state ? p->brightness : p->offbrightness;
+	switch (p->model) {
+		case HD44780_MODEL_WINSTAR_OLED:
+			cmd = WINST_MODESET | WINST_TEXTMODE \
+			    | (brightness >= MAX_BRIGHTNESS/2 ? WINST_PWRON : WINST_PWROFF);
+			p->hd44780_functions->senddata(p, 0, RS_INSTR, cmd);
+			break;
+
+		case HD44780_MODEL_PT6314_VFD:
+			cmd =  p->func_set_mode & ~PT6314_BRIGHT_MASK;
+			if (brightness >= 3 * MAX_BRIGHTNESS / 4)
+				cmd |= PT6314_BRIGHT_100; /* = 0x00 */
+			else if (brightness >= MAX_BRIGHTNESS / 2)
+				cmd |= PT6314_BRIGHT_75; /* = 0x01 */
+			else if (brightness > MAX_BRIGHTNESS / 4)
+				cmd |= PT6314_BRIGHT_50; /* = 0x02 */
+			else
+				cmd |= PT6314_BRIGHT_25; /* = 0x03 */
+			p->hd44780_functions->senddata(p, 0, RS_INSTR, cmd);
+			break;
+
+		default:
+			/* do nothing for other models */
+			break;
+	}
+}
+
+static void
+hd44780_set_backlight_config_cmds(PrivateData *p, int state)
+{
+	int brightness = state ? p->brightness : p->offbrightness;
+	int i;
+	unsigned char cmd;
+
+	if (brightness >= MAX_BRIGHTNESS/2) {
+		for (i=0; i<sizeof(p->backlight_cmd_on); i++) {
+			cmd =  (unsigned char)(p->backlight_cmd_on >> (24 - i*8)) & 0xff;
+			if (cmd)
+				p->hd44780_functions->senddata(p, 0, RS_INSTR, cmd);
+		}
+	}
+	else {
+		for (i=0; i<sizeof(p->backlight_cmd_off); i++) {
+			cmd =  (unsigned char)(p->backlight_cmd_off >> (24 - i*8)) & 0xff;
+			if (cmd)
+				p->hd44780_functions->senddata(p, 0, RS_INSTR, cmd);
+		}
+	}
+}
 
 /**
  * Turn the LCD backlight on or off.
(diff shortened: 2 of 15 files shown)


mskalski:
@haraldg: Please merge it to master, so I could add backlight handling as I described in #94.

In #94 I described how would configuration look like, here is I would do from programmers view:

There are many places which directly use `have_backlight` setting, which to this moment meant "using external pin".
From now it should mean, if nonzero, that display have some way of controlling backlight, but exact way is decribed using bits of this variable (0 = no backlight, 0x01 - external pin, 0x02 - internal commands, 0x04 - commands configured with option `BacklightCmdOn` and `BacklightCmdOff`, any combination could be virtually valid.

Now almost every connection type check something like 
``` C
if (p->have_backlight) {
    /* do something with backlight */
}
```

which should not be used this way as is little misleading.

I would change it do some inline function as for example:
``` C
if (have_backlight_pin(p) {
    /* do something with backlight */
}
```

and function would be
``` C
int have_backlight_pin(p) {
    return (p->backlight_mode & BACKLIGHT_EXTERNAL_PIN);
}
```
where `backlight_mode` is value read from configuration 'Backlight` setting.

And setting using internal commands would be:

``` C
void hd44780_set_backlight_internal(PrivateData *p, int state)
{
    unsigned char cmd;
    int brightness = state ? p->brightness : p->off_brightness;
    switch (p->model) {
        case HD44780_MODEL_WINSTAR_OLED:
            cmd = WINST_MODESET | WINST_TEXTMODE \
                 | (brightness >= 500 ? WINST_PWRON : WINST_PWROFF);
            p->hd44780_functions->senddata(p, 0, RS_INSTR, cmd);
            break;

        case HD44780_MODEL_PT6314_VFD:
            cmd =  p->func_set_mode;
            if (brightness >= 800)
                cmd |= PT6314_BRIGHT_100; /* = 0x00 */
            else if (brightness >= 500)
                cmd |= PT6314_BRIGHT_75; /* = 0x01 */
            else if (brightness > 300)
                cmd |= PT6314_BRIGHT_50; /* = 0x02 */
            else
                cmd |= PT6314_BRIGHT_25; /* = 0x03 */
            p->hd44780_functions->senddata(p, 0, RS_INSTR, cmd);
            break;

        default:
            /* do nothing */
            break;
    }
}
```

For PT6314_VFD we need to save mode used during initialization (`p->func_set_mode`). Proposed steps are 300, 500, 800, so for default settings of `Brightness` and `OffBrightness` would be fully on and minimal on.

Now main function for backlight in hd44780 driver would be:
``` C
MODULE_EXPORT void
HD44780_backlight(Driver *drvthis, int on)
{
	PrivateData *p = (PrivateData *) drvthis->private_data;

	/* Immediately return if noway of setting backlight is available
           or no change is necessary */
	if (!p->backlight_mode || p->backlightstate == on)
		return;
	if (p->hd44780_functions->backlight != NULL)
		p->hd44780_functions->backlight(p, on);

        if (p->backlight_mode & BACKLIGHT_INTERNAL)
            hd44780_set_backlight_internal(p, on);

        if (p->backlight_mode & BACKLIGHT_CONFIG_CMDS)
            hd44780_set_backlight_config_cmds(p, on);

	p->backlightstate = on;
}
```

And setting backlight using your (@haraldg) aproach would be:
``` C
void hd44780_set_backlight_config_cmds(PrivateData *p, int state)
{
    int brightness = state ? p->brightness : p->off_brightness;
    
    if (brightness >= 500 && p->brightness_cmd_on)
            p->hd44780_functions->senddata(p, 0, RS_INSTR, p->brightness_cmd_on);
    if (brightness < 500 && p->brightness_cmd_off)
            p->hd44780_functions->senddata(p, 0, RS_INSTR, p->brightness_cmd_off);
}
```

@haraldg, @trex2000, what do you think about it? Any suggestions are welcome.

trex2000:
Looks quite good. The only thing, I would use #define-s for the numerical constants (e.g. 500)


___________________________________________________________________
Until you spread your wings, you have no idea how far you can walk

mskalski:
I made initialization wrong (function `common_init()`) - I shouldn't change initialization order for standard and extended mode. I have to restore default init sequence and chage it only if model is WINSTAR_OLED.

server/drivers/hd44780-low.h
```diff
@@ -75,6 +75,28 @@
...
+#define HD44780_MODEL_DEFAULT		0
+/** Extended model - 4 lines activated with special bits and extra commands (EXTREG) */
+#define HD44780_MODEL_EXTENDED		1
+/** WINSTAR WEH00xxyyA (WEH001604A, WEH002004A, ...) and others, requires slightly different
+ * initialization sequence and have extra commands for handling brightness */
+#define HD44780_MODEL_WINSTAR_OLED	2
+/** PTC PT6314 VFD Displays - have extra command for setting four-level brightness.
+ *  Otherwise is compatible to other HD44780s */
+#define HD44780_MODEL_PT6314_VFD	3
+
```
haraldg, line 95: I think I would use here an enum, but don't care either way.

server/drivers/hd44780.c
```diff
@@ -112,6 +112,52 @@ unsigned char HD44780_scankeypad(PrivateData *p);
...
+
+	{ "winstar_oled", HD44780_MODEL_WINSTAR_OLED },
+	{ "weh00xxyya",   HD44780_MODEL_WINSTAR_OLED },
+	{ "winstar",      HD44780_MODEL_WINSTAR_OLED },
+	{ "winstaroled",  HD44780_MODEL_WINSTAR_OLED },
+
+	{ "pt6314_vfd",   HD44780_MODEL_PT6314_VFD },
+	{ "pt6314",       HD44780_MODEL_PT6314_VFD },
+	{ "pt6314vfd",    HD44780_MODEL_PT6314_VFD },
+
```
haraldg: Should we really have that many synonyms for winstar and pt6314? While it might be convenient, I think it can also cause confusion whether configurations are equivalent or not. At least we shouldn't have synonyms that aren't mentioned in the documentation. I recommend dropping them here.

haraldg (commented):
Looks mostly good to me. I don't see where you changed the init sequence, that you say you got wrong, however?

haraldg:
Your proposal looks good to me, however I didn't look into how much work this would be. But its your judgement anyway.

I don't care if the brightness values are #defined or not. I probably would write them as 1000/2 and use 1000/4 instead of 300 and 3*1000/4 instead of 800. And then maybe we have MAX_BRIGHTNESS definened as 1000 somewhere already? However I don't care in this case, as it is obvious enough what this is. (Much more then when feeding some magic number to the display controller.)

I probably won't merge this until people had some time to test it. Especially a tested-by from @trex2000 would be nice. However you can always push additional commits to this PR. As long as the changes are in different commits, we can pick them apart at any time easily. You can also overwrite this PR instead of opening a new one.

mskalski:
I broke with commit 7a631081b2726602bbc528958e0af6dc034a9499 initialization sequence for "Extended-mode" displays, fixed in 8135c5576007d46ba1a61108a04c1510b10e420b.

I thought using initialization sequence I found for Winstar OLED could be incorrect for standard ones, but in fact it is OK - I found exact init sequence in original HD44780 manual.

Also removed unnecessary/duplicate values for Winstar and VFD, just left one with generic WEH00xxyyA name.

Now making changes with backlight.

mskalski:
OK, I  think it is done.

@trex2000, Please check it if it works for your display.

trex2000:
Will do so these days. Thanks for the great work.


___________________________________________________________________
Until you spread your wings, you have no idea how far you can walk

mskalski:
I finally had some time to test my changes and found some bugs, fixed now.

Tested as previously using i2c connection for Winstar OLED (16x4), standard 16x4 and extended mode PC2004U (20x4). Tested mixed configurations and using internal commands from config file for Winstar OLED. 

Everything seems working now.

trex2000:
Hello,Today I've also compiled and tested it on the PT6314 VFD driver.I did not have a kodi installation, therefore brightness change from the client could not test, however I've tested the various brightness levels from the config file and it worked without problems. 

    On Friday, October 6, 2017, 9:19:46 AM GMT+3, mskalski <notifications@github.com> wrote:  
 
 
I finally had some time to test my changes and found some bugs, fixed now.

Tested as previously using i2c connection for Winstar OLED (16x4), standard 16x4 and extended mode PC2004U (20x4). Tested mixed configurations and using internal commands from config file for Winstar OLED.

Everything seems working now.

mskalski:
I usually test my display directly invoking commands from LCDproc developer guide (http://lcdproc.sourceforge.net/docs/current-dev.html#language-messages)
Usually I use `nc` and enter commands on stdin:

```
$ nc localhost 13666

hello
hello
screen_add test
widget_add test line1 string
widget_set test line1 1 1 "Hello world!"
screen_set test -backlight off
screen_set test -backlight on

```

And you should see something like this:
```
hello
connect LCDproc 0.5dev protocol 0.3 lcd wid 16 hgt 2 cellwid 5 cellhgt 8
screen_add test
success
listen test
widget_add test line1 string
success
widget_set test line1 1 1 "Hello world!"
success
screen_set test -backlight off
success
screen_set test -backlight on
success
```

mskalski:
@trex2000 what do you think about default on and off thresholds for your display?
 
Now default levels of `Brightness` and `OffBrightness` are 800 and 300, which means that PT6314 VFD display will be (by default, if not configured differently) respectively 100% and 50% of maximum brightness. Is it ok in your opinion?

server/drivers/hd44780.c
```diff
@@ -112,6 +112,164 @@ unsigned char HD44780_scankeypad(PrivateData *p);
...
+{
+	int result = 0;
+	int i;
+	int any = 0;
+
+	/* first find just whole option name */
+	for (i=0; i<sizeof(bl_value_mapping)/sizeof(bl_value_mapping[0]); i++) {
+		if (strcasecmp(value, bl_value_mapping[i].name) == 0)
+			return bl_value_mapping[i].value;
+	}
```
haraldg: Is there any reason for this loop? Maybe it's too late already, but isn't this covered by the following code anyway?

server/drivers/hd44780.c
```diff
@@ -112,6 +112,164 @@ unsigned char HD44780_scankeypad(PrivateData *p);
...
+
+	/* now find possibly all instances */
+	while (*value != '\0') {
+		const char *eot = strpbrk(value, ", |");
+		if (!eot)
+			eot = value + strlen(value);
+
+		i = 0;
```
haraldg: redundant code

server/drivers/hd44780.c
```diff
@@ -112,6 +112,164 @@ unsigned char HD44780_scankeypad(PrivateData *p);
...
+
+			/* break loop when marker found*/
+			if (bl_value_mapping[i].name[0] == '\0')
+				/* found unknown value name - so invalid format */
+				return -1;
+
+			len = strlen(bl_value_mapping[i].name);
+			if (len != eot - value)
+				continue;
+
```
haraldg: I think the call to strlen is unnecessary: both value and bl_value_mapping[i].name need to be 0 terminated anyway ...

server/drivers/hd44780.c
```diff
@@ -112,6 +112,164 @@ unsigned char HD44780_scankeypad(PrivateData *p);
...
+	if (any) {
+		return result;
+	}
+
+	return -1;
+}
+
+static const char *backlight_type_str(int backlight_type)
+{
+	static char buffer[256] = "";
```
haraldg: The lcdproc driver API disallows static (global and local) variables in drivers. Maybe it's easiest to just drop the logging code using this function. Or turn it into a debug output and just log the numeric value...

server/drivers/hd44780.c
```diff
@@ -466,23 +666,74 @@ HD44780_init(Driver *drvthis)
 void
 common_init(PrivateData *p, unsigned char if_bit)
 {
-	if (p->ext_mode) {
+	/* Set initial brightness according to Brightness setting.
+	 * This assumes that initially backlight is on (or if not used at all) */
```
haraldg: s/if/is/

server/drivers/hd44780.c
```diff
@@ -466,23 +666,74 @@ HD44780_init(Driver *drvthis)
...
+		p->hd44780_functions->senddata(p, 0, RS_INSTR, cmd_funcset | EXTREG);
 		p->hd44780_functions->uPause(p, 40);
 		p->hd44780_functions->senddata(p, 0, RS_INSTR, EXTMODESET | FOURLINE);
 		p->hd44780_functions->uPause(p, 40);
 	}
-	p->hd44780_functions->senddata(p, 0, RS_INSTR, FUNCSET | if_bit | TWOLINE | SMALLCHAR | p->font_bank);
+
+	if (p->model == HD44780_MODEL_PT6314_VFD) {
+		cmd_funcset &= ~PT6314_BRIGHT_MASK;
+		p->func_set_mode = cmd_funcset;
```
haraldg: This line seems redundant to the same line below ...

server/drivers/hd44780.c
```diff
@@ -877,6 +1128,67 @@ HD44780_set_brightness(Driver *drvthis, int state, int promille)
...
+	int i, shift;
+	unsigned char cmd;
+
+	/* Assume two levels of brightness */
+	if (brightness >= MAX_BRIGHTNESS/2) {
+		for (i=0; i<sizeof(p->backlight_cmd_on); i++) {
+			shift = (sizeof(p->backlight_cmd_on) - i - 1)*8;
+			cmd =  (unsigned char)(p->backlight_cmd_on >> shift) & 0xff;
+			if (cmd) {
+				fprintf(stderr, "hd44780_set_backlight_config_cmds: sending BL on %02x\n", cmd);
```
haraldg: Please no calls to fprintf(stderr, ...) - either use the standard report()/debug() logging or don't write anything at all.

server/drivers/hd44780.c
```diff
@@ -877,6 +1128,67 @@ HD44780_set_brightness(Driver *drvthis, int state, int promille)
...
+				p->hd44780_functions->senddata(p, 0, RS_INSTR, cmd);
+			}
+		}
+	}
+	else {
+		for (i=0; i<sizeof(p->backlight_cmd_off); i++) {
+			shift = (sizeof(p->backlight_cmd_on) - i - 1)*8;
+			cmd =  (unsigned char)(p->backlight_cmd_off >> shift) & 0xff;
+			if (cmd) {
+				fprintf(stderr, "hd44780_set_backlight_config_cmds: sending BL off %02x\n", cmd);
```
haraldg: ditto

haraldg (changes requested):
This looks pretty good. Most comments are minor nits, however the API issue is serious - even though it actually works with our current code.

trex2000:
Yes, you're right.I've never thought about using nc.The default off brightness in my opinion should be the lowest brightness level, but if it's not ok for the OLED, for example, I'm also fine with the 50%, not 25%.
 

    On Saturday, October 7, 2017, 1:27:15 AM GMT+3, haraldg <notifications@github.com> wrote:  
 
 
@haraldg requested changes on this pull request.

This looks pretty good. Most comments are minor nits, however the API issue is serious - even though it actually works with our current code.

In server/drivers/hd44780.c:
> +
+};
+
+/* parses null-terminated option value for backlight handling and returns bitmask of enabled values
+ * Returns -1 if value is not valid */
+static int parse_backlight_option( const char *value )
+{
+	int result = 0;
+	int i;
+	int any = 0;
+
+	/* first find just whole option name */
+	for (i=0; i<sizeof(bl_value_mapping)/sizeof(bl_value_mapping[0]); i++) {
+		if (strcasecmp(value, bl_value_mapping[i].name) == 0)
+			return bl_value_mapping[i].value;
+	}

Is there any reason for this loop? Maybe it's too late already, but isn't this covered by the following code anyway?

In server/drivers/hd44780.c:
> +	int i;
+	int any = 0;
+
+	/* first find just whole option name */
+	for (i=0; i<sizeof(bl_value_mapping)/sizeof(bl_value_mapping[0]); i++) {
+		if (strcasecmp(value, bl_value_mapping[i].name) == 0)
+			return bl_value_mapping[i].value;
+	}
+
+	/* now find possibly all instances */
+	while (*value != '\0') {
+		const char *eot = strpbrk(value, ", |");
+		if (!eot)
+			eot = value + strlen(value);
+
+		i = 0;

redundant code

In server/drivers/hd44780.c:
> +		if (!eot)
+			eot = value + strlen(value);
+
+		i = 0;
+		for (i=0; i<sizeof(bl_value_mapping)/sizeof(bl_value_mapping[0]); i++) {
+			size_t len;
+
+			/* break loop when marker found*/
+			if (bl_value_mapping[i].name[0] == '\0')
+				/* found unknown value name - so invalid format */
+				return -1;
+
+			len = strlen(bl_value_mapping[i].name);
+			if (len != eot - value)
+				continue;
+

I think the call to strlen is unnecessary: both value and bl_value_mapping[i].name need to be 0 terminated anyway ...

In server/drivers/hd44780.c:
> +		while (*value == ' ' || *value == '|' || *value == ',')
+			value ++;
+
+		/* try next one if present */
+	}
+
+	if (any) {
+		return result;
+	}
+
+	return -1;
+}
+
+static const char *backlight_type_str(int backlight_type)
+{
+	static char buffer[256] = "";

The lcdproc driver API disallows static (global and local) variables in drivers. Maybe it's easiest to just drop the logging code using this function. Or turn it into a debug output and just log the numeric value...

In server/drivers/hd44780.c:
> @@ -466,23 +666,74 @@ HD44780_init(Driver *drvthis)
 void
 common_init(PrivateData *p, unsigned char if_bit)
 {
-	if (p->ext_mode) {
+	/* Set initial brightness according to Brightness setting.
+	 * This assumes that initially backlight is on (or if not used at all) */

s/if/is/

In server/drivers/hd44780.c:
>  		p->hd44780_functions->uPause(p, 40);
 		p->hd44780_functions->senddata(p, 0, RS_INSTR, EXTMODESET | FOURLINE);
 		p->hd44780_functions->uPause(p, 40);
 	}
-	p->hd44780_functions->senddata(p, 0, RS_INSTR, FUNCSET | if_bit | TWOLINE | SMALLCHAR | p->font_bank);
+
+	if (p->model == HD44780_MODEL_PT6314_VFD) {
+		cmd_funcset &= ~PT6314_BRIGHT_MASK;
+		p->func_set_mode = cmd_funcset;

This line seems redundant to the same line below ...

In server/drivers/hd44780.c:
> +}
+
+static void
+hd44780_set_backlight_config_cmds(PrivateData *p, int state)
+{
+	int brightness = state ? p->brightness : p->offbrightness;
+	int i, shift;
+	unsigned char cmd;
+
+	/* Assume two levels of brightness */
+	if (brightness >= MAX_BRIGHTNESS/2) {
+		for (i=0; i<sizeof(p->backlight_cmd_on); i++) {
+			shift = (sizeof(p->backlight_cmd_on) - i - 1)*8;
+			cmd =  (unsigned char)(p->backlight_cmd_on >> shift) & 0xff;
+			if (cmd) {
+				fprintf(stderr, "hd44780_set_backlight_config_cmds: sending BL on %02x\n", cmd);

Please no calls to fprintf(stderr, ...) - either use the standard report()/debug() logging or don't write anything at all.

In server/drivers/hd44780.c:
> +	if (brightness >= MAX_BRIGHTNESS/2) {
+		for (i=0; i<sizeof(p->backlight_cmd_on); i++) {
+			shift = (sizeof(p->backlight_cmd_on) - i - 1)*8;
+			cmd =  (unsigned char)(p->backlight_cmd_on >> shift) & 0xff;
+			if (cmd) {
+				fprintf(stderr, "hd44780_set_backlight_config_cmds: sending BL on %02x\n", cmd);
+				p->hd44780_functions->senddata(p, 0, RS_INSTR, cmd);
+			}
+		}
+	}
+	else {
+		for (i=0; i<sizeof(p->backlight_cmd_off); i++) {
+			shift = (sizeof(p->backlight_cmd_on) - i - 1)*8;
+			cmd =  (unsigned char)(p->backlight_cmd_off >> shift) & 0xff;
+			if (cmd) {
+				fprintf(stderr, "hd44780_set_backlight_config_cmds: sending BL off %02x\n", cmd);

ditto

server/drivers/hd44780.c
mskalski: I want to specify `Backlight` option as a list separated with commas or `|` of following values:
`internal`, `external` and `internalCmds` and to allow other single values as `none` and all other boolean ones.

So first loop just checks single values (of all allowed), and followed loops search just combination of three first possibilities.

server/drivers/hd44780.c
mskalski: `strpbrk()` returns NULL if does not find searched character (here ` `, `|` and comma), and I want `eot` to point to first found character or end of string if none found.

server/drivers/hd44780.c
mskalski: Yes, `value` is 0-terminated, but I want to compare to first character other than comma, `|` and space. So comparison is done on only first `len` characters of `value`.

server/drivers/hd44780.c
mskalski: Yes, you are right. I just change getting textual of option value to report function and use buffer on stack.

server/drivers/hd44780.c
mskalski: You are right.

server/drivers/hd44780.c
mskalski: Yes, it is just left-over after debugging.

But I would like to leave some other way of debugging this (when for example report level is debug or more), but I don't see a way without passing `Driver *` structure pointer to this function which I want to avoid.

So I just remove this.

mskalski: Sorry, just noticed change in report() API - I thought that this is called through drvthis->report() pointer (as it was in 0.5.7 version). 

Now changed as you suggested.

mskalski:
@trex2000  If default off brightness at 50% is ok for you, I'll just leave it as is (it is simpler and more intuitive). I didn't see this VFD in action, so I cannot decide. On the other hand if you want to enable internal brightness handling, and so you have to edit LCDd.conf and changing off brightness is simple.

OLED has only two levels, so there is no choice for it :-)

mskalski:
@haraldg Fixed API issue and removed fprintf-s.

Option `Backlight` parsing was left as was, I commented why I did it that way in code. Basically it is because I want to allow list of few values and single other values.

server/drivers/hd44780.c
haraldg: Ah, now I see: Your special marker to end the loop early. Well, this code is way to smart for anybody to maintain in the long term ...

How about just parsing the actually meaningful values, and if that fails then falling back to parsing again with conf_get_bool() in the init function? Should be much clearer what's going on that way, I hope.

server/drivers/hd44780.c
haraldg: Actually I was refereing to the `i = 0;` line.

server/drivers/hd44780.c
haraldg: I guess I confused strpbrk() with strtok_r(), but the strlen() still looks unnecessary to me: `strncasecmp(value, bl_value_mapping[i].name, eot - value)` should do the trick?

server/drivers/hd44780.c
```diff
@@ -238,18 +238,13 @@ static int parse_backlight_option( const char *value )
 	return -1;
 }
 
-static const char *backlight_type_str(int backlight_type)
+/* reports textually setting of backlight */
+static void report_backlight_type(int report_level, int backlight_type)
 {
-	static char buffer[256] = "";
+	/* should be enough for current options set and all possible combinations */
+	char buffer[256] = "";
```
haraldg: While this is certainly true, it is very easy to miss, when adding new options. So at least we should check against overflow in the code below.

server/drivers/hd44780.c
```diff
@@ -258,7 +253,6 @@ static const char *backlight_type_str(int backlight_type)
 			strcat(buffer, bl_value_mapping[i].name);
 			unknown &= ~bl_value_mapping[i].value;
```
haraldg: Actually you want `unkown |= bl_value_mapping[i].value;`?

haraldg (changes requested):
I know I'm being quite strict here, but then it's just a question of time until somebody runs the code through a source code analysis suite and starts opening issues about such things ...

server/drivers/hd44780.c
mskalski: No, it's OK, just `unknown` shall be initialized to `backlight_type`. I meant to print unknown bits if there are any.

mskalski:
@haraldg Fixed according to your suggestions.

I didn't separate reading "new" values for option `Backlight` though, just added comments to function `parse_backlight_option()`. I think it's better to leave reading **all recognized** options in this function, and not call `config_get_bool()` in `HD44780_init()` in case of failure.

But if you think it would be better, I can add in `HD44780_init()`:

``` C
	/* parse backlight option. Default is model specific */
	s			= p->model ==  (HD44780_MODEL_WINSTAR_OLED || p->model == HD44780_MODEL_PT6314_VFD) ?
				               "internal" : "none";
	s			= drvthis->config_get_string(drvthis->name, "backlight", 0, s);
	p->backlight_type	= parse_backlight_option(s);
+	if (p->backlight_type < 0) {
+		/* invalid value - try to parse as boolean, with default value also of -1 */
+		tmp = drvthis->config_get_bool(drvthis->name, "backlight", 0, -1);
+		if (tmp < 0)
+			p->backlight_type = -1;
+		else
+			p->backlight_type = tmp ? BACKLIGHT_EXTERNAL_PIN : BACKLIGHT_NONE;
+	}

	if (p->backlight_type < 0) {
		report(RPT_ERR, "%s: unknown Backlight type: %s", drvthis->name, s);
		return -1;
	}
```

mskalski:
Also changed debug texts written for `internalCmds` backlight  handling and also added some `report(RPT_DEBUG)` for `internal` handling.

server/drivers/hd44780.c
```diff
@@ -238,30 +243,62 @@ static int parse_backlight_option( const char *value )
...
+
+	size_t dlen = strlen(dst);
+	size_t slen = strlen(src);
+
+	if (slen + dlen < dsize) {
+		memcpy(dst + dlen, src, slen);
+		dst[dlen+slen] = 0;
+	}
+	else if (dlen < dsize) {
+		memcpy(dst + dlen, src, slen - (dsize - dlen));
```
haraldg: I think here you just want `dsize - dlen` instead of `slen - (dsize - dlen)` ...

haraldg (commented):
If you actually try to explain your reasoning, why manually parsing for true/false values is better, you might be able to convince me. ATM it looks to me like it is more lines of code and more complicated and less a representation of the idea behind the algorithm then the alternative you are proposing. So I can't quite see why you think that your solution is better then your alternativ solution ... :)

server/drivers/hd44780.c
mskalski: Yes, you are right

mskalski: I'm not saying that manually parsing true/false values is better, but it costs nothing in this case, as  that complication of function `parse_backlight_option()` is because of handling list of options and assumption that only few of them can be combined (first three) and `none` shall be single one (as this looks natural). 

It would be only little simpler if it would be allowed to add `none` to list, as it is harmless (value is 0).

mskalski: It's just adds few records to table `bl_value_mapping` with boolean aliases.

On the other hand I think as there is specialized function for parsing value, it should handle all possible values, even if it doesn't look simple.

mskalski:
@haraldg Sorry, I put my comment in wrong place, so here once again:

I'm not saying that manually parsing true/false values is better, but it costs nothing in this case, as this complication of function `parse_backlight_option()` is because of handling list of options and assumption that only few of them can be combined (first three) and `none` shall be single one (as this looks natural).

It would be only little simpler if it would be allowed to add `none` to list, as it is harmless (value is 0)
Supporting true/false values in this function is matter of adding few records of  boolean aliases to table `bl_value_mapping`.

On the other hand I think as there is specialized function for parsing value, it should handle all possible values, even if it doesn't look simple and don't do the same in other place.

server/drivers/hd44780.c
haraldg: > It's just adds few records to table bl_value_mapping with boolean aliases.

Yes, having redundant definition of the boolean values is one of the downsides IMO.

> On the other hand I think as there is specialized function for parsing value, it should handle all possible values, even if it doesn't look simple and don't do the same in other place.

Here I fail to see the rationale. Can you explain this to me?

As I see it the boolean values are basically deprecated and only supported for backwards compatibility. I think it is straigt forward to have one function parsing the real thing and the backwards compatibility fallback at the call site.

I don't have a strong opinion on this, but I really would like to understand the merit you see before merging something that feels counter-intuitive and complicated to me.

mskalski:
@haraldg Moved handling boolean deprecated values out of `parse_backlight_option()` function.

As you can see in last commit that `parse_backlight_option()` is not any simpler, as most of code there handles possible and impossible combinations of values. 

And to handle backward compatibility I had to add code to `HD44780_init()` function.

This is what I meant in last comment. Sorry for my English, I'm not native-English speaker, I have sometimes problems with expressing what I really mean.

mskalski:
And to make function `parse_backlight_option()` simpler, we would have to give up recognizing possible and invalid combinations of values.

@haraldg Or maybe You have some idea, how would be this done?

mskalski:
Or maybe it would good to make such general form of handling maps of names to integers or lists of values to shared dedicated functions, for example in `shared/str.c`?

haraldg:
Already 3 lines of code saved, yay ... ;)

Seriously, again: Maybe I still haven't understood how parse_backlight_option() is supposed to work, but I would think that now the first for-loop can be dropped without a change in behaviour?

I felt tempted to suggest moving code into `shared/...` several times, but configuration handling is going to be completely reworked in the mid term anyway. Maybe strappend() should be moved to `shared/str.c` but then doing it right would probably involve using strlcat() and providing our own implementation on platforms that don't have it. I don't want to nudge you into working on our build system on top of everything else you did already, so strappend() is fine for now.

mskalski:
OK, I'll try explain again :-)

First I want to allow combinations of some values (namely `external`, `internal`, `internalCmds`) **and not allow** to combine it with `none`.

So to achieve this, I did it with algorithm below:
-   First loop searches exact option value (single one) and if it matches returns associated integer mask (or 0 on `none`). This loop goes for **all** elements of `bl_value_mapping`, as every element of this table is valid as single value.

-   Next 2 nested loops work as below:
    -   Find next token delimiter (`','`, `' '` and `'|'`), if not found, get token to the end of string. 
    -   Next in nested loop compare with subsequent elements from table `bl_value_mapping` till we don't match our special marker. If found, we tried before all allowed for combining values and it did not match.
        So for this moment we can return -1 as all valid for combination elements have been checked.

Without first loop we need to know in second loop if matching element is allowed to combine (which is simple, we can mark them with third member of `struct bl_value_mapping`), also we need to know if there were other elements for combination (this is simple too, we can have extra variable telling that) but  also need to know if there will be next one element on list (which is harder, though possible of course). 

So we can achieve the same effect if we have extra field in `struct bl_value_mapping` and two extra variables in second loop, which is not worth of.

Alternatively we may just don't care about allowed and not allowed combinations of values, so first loop is not necessary, and special marker is unnecessary also.

haraldg:
Thanks for the explanation, so I try to paraphrase, tell me if I'm right
or not:

The first loop is still necessary to check if 'none' was passed as
backlight option without allowing it to be combined with other options.

If that's true, then a simple

if (strcmp(value, "none")) ...

at the beginning should do the trick.

However the question whether combining 'none' with other values is valid
is a bit philosophical. So from a practical point of view, I just wouldn't
bother.

Actually, from a keep it simple and stupid point of view one might even
wonder, why 'none' needs to be a valid value when the option can be
omitted altogether anyway. Though I concede that from a user interface
point of view such a value has merit.

Which however makes me realize something I really should have
pointed out when reviewing this for the first time. (Sorry, I'm reviewing
your code in the breaks between paid work and concentration slips sometimes.)

The thing is, there is already an established syntax for "combining" multiple
values in lcdproc: Just repeat the option multiple times. For the old
backlight option, where the value was a simple bool, there was no point in
checking the option on multiple lines, so only the first instance was ever
checked. But the conf_get_* functions do have an interator argument exactly
to allow handling such cases.

mskalski:
Yes, I forgot about syntax involving multiple times the same option. Maybe I am used to semantics of configuration files where last (or first) occurrence of given option in given section "wins". 

In fact I was confused during my tests when I changed option `Backlight` for tests and it did nothing - always appeared as value 'yes' :-).

OK, so summarizing - I shall just replace both loops with simple iterating through multiple occurrences of option `Backlight`, yes? This has disadvantage of detecting a mix of old boolean and new "proper" names a bit difficult. But OK.

haraldg:
mskalski writes:
> Maybe I am used to semantics of configuration files where last (or first) occurrence of given option in given section "wins". 

This is what lcdproc does "most of the time", but not always ...

Yes, our code base is a huge mess.
 
> OK, so summarizing - I shall just replace both loops with simple iterating through multiple occurrences of option `Backlight`, yes?

If I understood you correct, then it seems reasonable.

Maybe just loop over all occurrences of option `Backlight` directly in
the init function and call some helper function with private data and the
configuration value there and setup flags and whatever in the helper function.
Maybe a helper function isn't even necessary?

But enough micro management. You did a good job so far, I'm sure you will
come up with something reasonable.

> This has disadvantage of detecting a mix of old boolean and new "proper" names a bit difficult. But OK.

Well, I guess people could disagree about what is the "proper" backwards
compatible behaviour in this case anyway. Let's just ignore this old-new-mix
issue and look forward to deprecating the old way in a few years from now.

Backwards compatiblity is for working configurations to keep working, not
for broken configurations like

Backlight=yes
Backlight=no

to stay bug compatible forever.

Maybe the switch to libelektra provides a nice excuse for dropping this
cruft sooner or maybe they can even deal with such a mess in their
configuration specification without any fancy workarounds in the code.
We will see.

mskalski:
@haraldg Fixed according to Your suggestions, now enabling use of multiple occurrences of option `Backlight` in configuration file with only first occurrence allowed to be boolean.

Also added warning about deprecated boolean syntax of this option.

Now is indeed much simpler.

mskalski:
@haraldg Regarding configuration, one more question: is it possible to have two sections with configuration for same type of driver, for example to show the same duplicated data on two HD44780 displays synchronized together?

This is possible for two different type of displays (ie hd44780 and curses).

haraldg:
Looks good to me. I still feel you are wasting way too many LoC just to
prevent people from shooting themselves in the foot, but that's a matter
of taste not reason, so I won't argue about it.

I will merge this (probably squashing some commits to make the history less
confusing) unless somebody speaks up soon. Since you are new to lcdproc
development, maybe you want to add a final commit to this PR to add yourself
to `CREDITS.md` ...

> Regarding configuration, one more question: is it possible to have two sections with configuration for same type of driver, for example to show the same duplicated data on two HD44780 displays synchronized together?

Interesting question, I don't know. I think it is not specified ATM, so
what ever the answer, it might change in the future at any time.

I guess it overstreches our current configuration syntax (I think I remember
something about ini combining multiple sections of the same name into one),
but in the long term I want to add new configuration syntax (sections
describing setups instead of drivers), which as a side effect would also
allow this to work if it doesn't already.

Do you have anything specific in mind?

Thanks a lot for your patient work!

trex2000:
Guys, I would like to thank you for your great work!
From my initial modification for the VFD driver, to just ‘make it work’ , this one is indeed one nice update!


___________________________________________________________________
Until you spread your wings, you have no idea how far you can walk

mskalski:
Thanks, @haraldg, @trex2000 

@haraldg I'm not sure if it is real preventing people shooting themselves in the foot, I just found that many times errors in configuration which were not detected at start, appeared later much more difficult to diagnose. So few lines of code can save much time of debugging later.

Thanks again, guys

trex2000:
All the best,
Regards, Barna Pető


___________________________________________________________________
Until you spread your wings, you have no idea how far you can walk

haraldg:
Merged this branch manually to squash a few commits together and cleaning up some whitespace issues. Thus closing this PR.

PR #96, opened 2017-10-07
Enables the Futaba TOSD-5711BB LED display to display the time with a flashing colon (:) between the hours & minutes and (optionally) seconds. 

the-black-eagle:


mskalski (commented):
I would rather not change indentation. It's difficult to see what have changed

the-black-eagle:
I ran it through 'indent' with the config file specified in the docs.  Basically, there is some extra processing in 'futaba_send_string' to determine if we want to display a time, a separate thread to actually flash the colon(s) and the addition of 4 flags to the driver's private data.

haraldg:
I agree with @mskalski - this PR is almost impossible to review. I recommend creating a new branch (based on master), doing the indent there again and just commiting the indent changes. (Add the exact options used to the commit message, so that it's easy to verify later.) Then rebase or merge your PR to the new branch. This should give you a commit with just the real changes without any manual editing. Then push -f your new two commit branch to this PR.

the-black-eagle:
Yeah, didn't realise that running indent on it would make it much harder for you guys.  I'll close this and re-submit a PR without the indent changes.

PR #97, opened 2017-10-07
Add support for displaying ':' between numbers when showing the time

the-black-eagle:
Flashes one of the colon icons on the display approx. once per second when displaying a time in the format HH:MM.   Also can flash two icons if displaying a time in the format HH:MM:SS


server/drivers/futaba.c
```diff
@@ -118,6 +126,65 @@ MODULE_EXPORT int stay_in_foreground = 0;
...
+            my_report.param1 = 0x02;
+            my_report.type.sym.count = 1;
+            my_report.type.sym.symbol[0].symName = 0x34;
+            my_report.type.sym.symbol[0].state = FUTABA_SYM_OFF;
+            futaba_send_report(p->my_handle, &my_report);
+            p->flash_state = 0;
+            sleep(1);
+        }
+    }
+
```
mskalski: When none of options `p->flash` nor `p->flash_state` will be set, thread will consume 100% of processor, as there is no sleep at all.

the-black-eagle: Indeed.  However, as it effectively does nothing in the loop it doesn't appear to affect the machine at all.  It has been running on my rig continuously for 5 days 24/7 and with a passively cooled pentium4, max temp has been 40C only (Max is 115C).  Note that Kodi is also running continuously.

I can't see a different way of blinking those icons within the driver without using a separate thread to do it.  If there is a better way of doing it, I'd be pleased to implement it.  BTW, lis driver also implements a separate thread (which reads not writes) but has no sleep() in it at all.

haraldg: The reason why this works in lis is because it uses a blocking call IIRC. Here I don't see a blocking call but I guess on your setup just is so that a flag is set and you have a sleep() in your tests most of the time.

That lis pulls in libpthread is a bit annoing but it is already there, so little point in trying to avoid it here. However it might be possible to implement blinking without a extra thread and at the same time fix some other issues with the code. I'm writing a proper review about that ...

server/drivers/futaba.c
```diff
@@ -174,16 +241,52 @@ futaba_send_string(Driver *drvthis)
...
+            p->flash = 1;   /* signal to thread_timer that we need
+                     * to start blinking the colon */
+            p->long_time = 0;
+        }       /* Only blink the one colon */
+
+        /* The following block is for a time in the format HH:MM:SS
+         * (Note that extra processing is needed in Kodi's xbmc
+         * lcdproc addon for this to work!)
+         * addon available at https://github.com/the-black-eagle/script.xbmc.lcdproc */
+
```
haraldg: I didn't look at it but this sounds like some very ugly hack. Is there no better solution?

server/drivers/futaba.c
```diff
@@ -527,11 +633,19 @@ futaba_init(Driver *drvthis)
...
+    	else {
+        p->thread_timer_var = thread_timer_var;
+        report(RPT_INFO, "[%s] Successfully created timer thread", drvthis->name);
+    	}
 
 	/* write to the file */
 	futaba_start_driver(drvthis);
 	memset(p->framebuf, ' ', p->width * p->height);
 	memcpy(p->old_framebuf, p->framebuf, p->width * p->height);
+	p->driver_init_complete = 1;
```
haraldg: Why not just create the thread at this point and drop the driver_init_complete flag?

haraldg (commented):
Interesting approach. I too wrote a driver for a display with additional dots and know that such a beast is difficult to integrate with lcdproc properly. I guess one of the problems you are trying to solve, is that the display has a fixed length in lcdproc but actually can display strings of different length depending on how many extra symbols it can use? Actually I changed the rendering of string widgets to allow "multibyte" characters to work around this. However other widgets are still left to do.

About your flashing implementation: Since the thread and the main programm are only synchronised every second, there will be a delay of up to a second, when LCDd rotates screens to change the colon as well... it would be a lot better, if the futaba_send_string() function could also update the extra symbols.

Actually, I think it can. As LCDd works ATM it redraws the screen on every frame. That is: By default your futaba_send_string() is called 8 times per second. Just call time() inside the function and display the colon on even seconds and don't on odd seconds. Should be quite easy.

Did you consider linking the colon flashing to the hearbeat setting?

PR #101, opened 2018-04-22
Rework serialPOS and add support for new displays

shenghaoyang:
Reworked serialPOS core to split protocols into their own source files.

Core changes:

- Added support for more driver API calls, it's up to the various protocol 
  implementations to support them
- Fixed some ``vbar`` / ``hbar`` segmentation issues

Added support for new protocols:

- cd5220
- EMAX through AEDEX
- UTC / S (Ultimate) through Logic Controls

Improved functionality for:

- ESC / POS
- Logic controls 

First time submitting a pull request, comments welcome.

TODO: rework input function for reading from keyboards attached to POS displays - no hardware to test.

The display used to test these commits seems to have a flawed Logic Controls command set emulation. If anyone has a Logic Controls display, and could give this a test, that would be welcome - the brightness control command is ignored by the test display (KPD220V7, ``http://kinghigh.com.cn/page88.html?product_id=54``)


haraldg:
Hi!

Thanks for working on this driver. This seems to be a major face lift!

Unfortunately your pull request is somewhat difficult to review as is,
so I haven't been able to read everything yet.


A few general comments:

I recommend you to rename .git/hooks/pre-commit.sample to
.git/hooks/pre-commit in your lcdproc git repository. This should warn
you about some of the more common formatting errors.

Please follow to coding style (as described in the developer guide).
In particular it is important to use one tab (8 columns) per indentation
level.

I see you use basically one commit per file changed. This is not ideal:
Among other issues it makes it very difficult to keep track of changes
when reviewing the code. Ideally you would have one commit per logical/
functional change. Some guidelines about this:
 * each commit must be able to compile
 * when code is moved between two files, both files need to be updated
   in the same commit
 * whitespace changes should be in separate commits from functional changes
 * documentation should be updated in the same commit that changes the
   behaviour
(I usually let people get away with violateing the last two rules, but
since some pretty deep reformatting of the pull request is necessary
anyway, maybe you can address this too.)

Functions not specific to the driver (like timespec_to_ns, retry_close)
should be moved to some general file - but I think most of them aren't
necessary at all and can just be dropped.

A lot of your changes are about error checking and reporting. A few
comments (which might not apply everywhere, common sense necessary)
about this:
* debug messages should be reported with debug() instead of report()
  (to omit them from production builds)
* messages with high priority (ie errors) should only be reported during
  initialization or shutdown - otherwise we risk spamming the syslog
  with 8 (or more) messages per second and fill up the log storage.
* many system calls that in theory can fail for historic reasons, usually
  don't fail on modern systems. For example on systems with virtual
  memory malloc() always succeeds except for blunt errors or exotic
  corner cases like when one is using memory locking. assert() is a good
  way to check such syscalls, though in many cases I don't bother
  checking return values at all.
* correct error handling often is difficult. If we can't (or won't) do
  it right, often it is best to not bother at all.


Specific comments:

retry_close(): This is not necessary (we rarely use signals, if we use signals,
we use SA_RESTART anyway) and retrying close() is actually wrong on
many operating systems including linux.

serialPOS_write(): While this function has the serialPOS_ prefix, it is
actually a generic function. Anyway: What's the reason for including this?
Are there known issues with display locking up and makeing write() block
indefinitely? I'm not fond of adding such a complex function without
a strong reason - and from looking at how the function is used, I can't
figure out what the reason is ... in particular I see that it is mostly
called with a timeout of 125ms, which is pretty short for a "let's not
hang indefinitely" timeout. I can also guess that there might be issues,
with displays at low baud rates where LCDd tries to send data faster then
the links supports, which in the case of blocking IO might cause
bad behaviour if the kernel does too much buffering. However this
function doesn't seem to do rate limiting either, so what is the rationale
behind it?

I think there is no point in providing stub implementations for
unsupported protocols.

The interface between the driver and the protocols seems rather complicated
to me. Maybe this can be simplified? Some ideas:
 * Do we really need private data of the protocol?
 * The size of the command buffer could be calculated in init once.
 * After your changes p->emulation_mode is plain unnecessary.

I think with your code if either Cellwid or Cellhgt is unspecified a
warning is printed and the default value is used for both. Not sure, if
this is intended?

regarding:
> /**
> * Maximum height of an ESC/POS display - arbitrary limit based on
> * the displays actually observed.
> */
> static const uint8_t EPSON_MAX_HEIGHT = 4;

I'm fine with imposing limits if they help to simplify the code. But I
don't see the point of arbitrary limits.

Thanks for all the work you are putting into this!
Harald

shenghaoyang:
Hey, 

Thanks for the comments.

> I see you use basically one commit per file changed. This is not ideal:
> Among other issues it makes it very difficult to keep track of changes
> when reviewing the code. Ideally you would have one commit per logical/
> functional change. Some guidelines about this:

Sure, I'll try to reorganize the commits instead of doling them out by files.

> debug messages should be reported with debug() instead of report()
> (to omit them from production builds)

Alright, didn't realize there was a difference between the two, since it 
seemed to be used interchangeably in the original driver. I did see it 
mentioned in the documentation, but wasn't really sure on which to follow.

> messages with high priority (ie errors) should only be reported during
> initialization or shutdown - otherwise we risk spamming the syslog
> with 8 (or more) messages per second and fill up the log storage.

Might it be a good idea to document this somewhere? Didn't figure this 
was a design goal.

With journald  running on systems I usually don't bother much about spamming 
syslog, since different units can be easily isolated. (Though this might be 
an issue, I admit, if we're running on something like LEDE) 

> retry_close(): This is not necessary (we rarely use signals, if we use signals,
> we use SA_RESTART anyway) and retrying close() is actually wrong on
> many operating systems including linux.

Sure. That saves a lot of troublesome code in handling EINTR. 
Original idea was trying to stick to the ``POSIX`` spec.

> serialPOS_write(): While this function has the serialPOS_ prefix, it is
> actually a generic function. Anyway: What's the reason for including this?
> Are there known issues with display locking up and makeing write() block
> indefinitely? I'm not fond of adding such a complex function without
> a strong reason - and from looking at how the function is used, I can't
> figure out what the reason is ... in particular I see that it is mostly
> called with a timeout of 125ms, which is pretty short for a "let's not
> hang indefinitely" timeout. I can also guess that there might be issues,
> with displays at low baud rates where LCDd tries to send data faster then
> the links supports, which in the case of blocking IO might cause
> bad behaviour if the kernel does too much buffering. However this
> function doesn't seem to do rate limiting either, so what is the rationale
> behind it?

Half the reason for it was to handle ``EINTR``. If the system will automatically
restart calls, then it can be slimmed down quite a bit. 

There is also a need to wrap ``write()`` to retry calls - 
if for some reason the whole buffer passed to write was not 
written (i.e. FIFO in serial device is full, and we're waiting for it to drain)

Given the amount of delta updates being performed, missing a chunk of data,
say, through an incomplete ``write()`` call might lead to the display's state 
being inconsistent with what the protocol driver expects it to be - and this
could lead to quite a mess on the physical screen. 

Though, of course, we could track this failure and request a full framebuffer
flush from the protocol drivers and have the whole screen state flushed on the 
next update cycle, I thought it would've added quite a bit of clutter to the 
already complicated interface between the main driver and the protocol driver - 
I'd much rather try to handle it in the main driver than requiring more 
functionality from the protocol drivers.

Besides, some of the commands expect a certain fixed-length 
sequence, - see the ``CURSOR_MOVE`` commands,
and it's impossible to restart a new screen flush halfway through. 
Putting received data into a FIFO managed by the main driver might work, 
but then it'd end up simply doing the same thing as a retried ``write()``.

Though, with knowledge of the command one could work around this 
and attempt to complete the write, but I think that would simply over-complicate
what was meant to be a simple driver.

Regarding the timeouts - wanted to have them included just in case 
there were slow displays and I didn't want to hold up the entire process, though
it's true there aren't many of those low-baud displays out there. If slowing 
the process down is acceptable then we can do away with the timeout. 

> I think there is no point in providing stub implementations for
> unsupported protocols.

Sure - will remove those.

> Do we really need private data of the protocol?
> The size of the command buffer could be calculated in init once.
> After your changes p->emulation_mode is plain unnecessary.

Thought some private data was necessary for the protocols themselves to store 
their own information in formats suitable for the protocol, if they need
to track the display's state. This could be included in the main private data 
structure that the driver uses, but I think it'd add unnecessary clutter to the 
main structure - the ``AEDEX`` protocol has no state to track, while when using 
the ``ESC / POS`` or ``CD5220`` protocols the driver tracks the state of the cursor 
and brightness of the display. Eventually I'd like them to track the custom characters
loaded into the display - something that would *certainly* not be required for 
protocols like ``AEDEX``.

> I think with your code if either Cellwid or Cellhgt is unspecified a
> warning is printed and the default value is used for both. Not sure, if
> this is intended?

This is intended. These configuration fields weren't included in the 
original driver, and requiring them would break existing installations.

I believe many used the ``AEDEX`` protocol, which wouldn't require these
fields anyway. 

Same goes for the ``Custom_chars`` field, which defaults to zero if not set.

> I'm fine with imposing limits if they help to simplify the code. But I
> don't see the point of arbitrary limits.

I wish I could impose hard limits according to the protocol, but there 
aren't clear specifications regarding what is allowed and documentation
for various display models suggest differing limits for display dimensions,
at least according to argument limits for the cursor-shift ``ESC / POS``
commands documented - I don't want to shift the cursor, if requested,
beyond what is allowed by the specification.

If it's desired to cope with wildcard displays that have sizes beyond what
is commonly observed out "in the field", then we could remove this limit, 
and go with the main driver limit.

Best, 

Shenghao Yang

haraldg:
Shenghao Yang writes:
> > messages with high priority (ie errors) should only be reported during
> > initialization or shutdown - otherwise we risk spamming the syslog
> > with 8 (or more) messages per second and fill up the log storage.
> 
> Might it be a good idea to document this somewhere? Didn't figure this 
> was a design goal.

Feel free to document it, if you have an idea where. But it isn't some
special design goal - just me thinking this is common sense.

> With journald  running on systems I usually don't bother much about spamming 
> syslog, since different units can be easily isolated. (Though this might be 
> an issue, I admit, if we're running on something like LEDE) 

Yes LEDE is one concern. (It is currently the linux distribution with the
most timely package updates after a new release, I think.)

The other is: At 8 lines/second we are spamming syslog with about 1kB/s -
even with logrotate that does make for a lot of data ...
 
> > retry_close(): This is not necessary (we rarely use signals, if we use signals,
> > we use SA_RESTART anyway) and retrying close() is actually wrong on
> > many operating systems including linux.
> 
> Sure. That saves a lot of troublesome code in handling EINTR. 
> Original idea was trying to stick to the ``POSIX`` spec.

Yeah, unfortunately this is one of the cases where posix is just broken.
Linux conforms to posix and HP-UX confrom to posix, but you can't write
code that works for both without #ifdefs. But fortunately this is a
historic problem that is not relevent since the introduction of SA_RESTART.

> > serialPOS_write(): 
>
> Given the amount of delta updates being performed, missing a chunk of data,
> say, through an incomplete ``write()`` call might lead to the display's state 
> being inconsistent with what the protocol driver expects it to be - and this
> could lead to quite a mess on the physical screen. 

Okay, I kind of see your concern, though I still suspect that the code
doesn't actually help solving it (or rather: it probably is worse then
classical blocking IO, which only causes issues in the case of EINTR).

I suggest you ignore this issue for now and focus on the multi protocol
support. We can discuss how to improve IO reliability later.

(Some ideas: Use stream based io using fdopen(), add a reliable io facility
to the server core and use that, decide that there is no real issue anyway,
stop using non-terminating signals, ...)
 
> Regarding the timeouts - wanted to have them included just in case 
> there were slow displays and I didn't want to hold up the entire process, though
> it's true there aren't many of those low-baud displays out there. If slowing 
> the process down is acceptable then we can do away with the timeout. 

I think it is actually necessary to slow the main process down, or
you are sabotaging your reliability goal from before.

> > Do we really need private data of the protocol?
> > The size of the command buffer could be calculated in init once.
> > After your changes p->emulation_mode is plain unnecessary.
> 
> Thought some private data was necessary for the protocols themselves to store 
> their own information in formats suitable for the protocol, if they need
> to track the display's state. This could be included in the main private data 
> structure that the driver uses, but I think it'd add unnecessary clutter to the 
> main structure - the ``AEDEX`` protocol has no state to track, while when using 
> the ``ESC / POS`` or ``CD5220`` protocols the driver tracks the state of the cursor 
> and brightness of the display. Eventually I'd like them to track the custom characters
> loaded into the display - something that would *certainly* not be required for 
> protocols like ``AEDEX``.

I don't think there is anything wrong with having some features not supported
by some protocols. Unless each protocol needs a vastly different state,
I think it is fine to have state variables for all possible features in
the main private data. - Especially since all protocols are statically linked
into the main driver anyway.

> > I think with your code if either Cellwid or Cellhgt is unspecified a
> > warning is printed and the default value is used for both. Not sure, if
> > this is intended?
> 
> This is intended. These configuration fields weren't included in the 
> original driver, and requiring them would break existing installations.

I think you misunderstood me: My point is that with the code as it is now
the user can't specify Cellhgt in the configuration file but leave
Cellwid at the default value, because the code will use default values
for both. (No big issue, just wonderin if this is intended.)
 
> I believe many used the ``AEDEX`` protocol, which wouldn't require these
> fields anyway. 

Then maybe only parse them in the protocol specific init functions, where
they are actually needed?

> Same goes for the ``Custom_chars`` field, which defaults to zero if not set.
> 
> > I'm fine with imposing limits if they help to simplify the code. But I
> > don't see the point of arbitrary limits.
> 
> I wish I could impose hard limits according to the protocol, but there 
> aren't clear specifications regarding what is allowed and documentation
> for various display models suggest differing limits for display dimensions,
> at least according to argument limits for the cursor-shift ``ESC / POS``
> commands documented - I don't want to shift the cursor, if requested,
> beyond what is allowed by the specification.

Hm, is there something bad going to happen?

Bad as in: It might overwrite some kind of firmware and brick the display.
If it just causes garbage to be displayed, that's ok: Anybody putting
garbage in there configuration should expect garbage on their display.
 
> If it's desired to cope with wildcard displays that have sizes beyond what
> is commonly observed out "in the field", then we could remove this limit, 
> and go with the main driver limit.

It's just a minor concern, but if the specification allows vendor foo
to produce a display with 5 lines then I see little point in preventing
it. And given how long it takes for a new patch to lcdproc to end up in
distributions like debian stable, it's not all that easy to lift the
limit after the fact.

Thanks for your consideration,
Harald

-- 
If you want to support my work:
see http://friends.ccbib.org/harald/supporting/
or donate via CLAM to xASPBtezLNqj4cUe8MT5nZjthRSEjrRQXN
or via peercoin to P98LRdhit3gZbHDBe7ta5jtXrMJUms4p7w

shenghaoyang:
Hi, 

Thanks for the comments.

 > serialPOS_write():

> >  Given the amount of delta updates being performed, missing a chunk of data,
> >  say, through an incomplete ``write()`` call might lead to the display's state
> >  being inconsistent with what the protocol driver expects it to be - and this
> >  could lead to quite a mess on the physical screen.
> 
> I suggest you ignore this issue for now and focus on the multi protocol
> support. We can discuss how to improve IO reliability later.

Sure. We can leave the original implementation as-is and work on the more 
interesting stuff.

> I don't think there is anything wrong with having some features not supported
> by some protocols. Unless each protocol needs a vastly different state,
> I think it is fine to have state variables for all possible features in
> the main private data. - Especially since all protocols are statically linked
> into the main driver anyway.

If we're going in this direction then I'd like to define a few more structures
and include those in the ``PrivateData`` structure for the driver - else there'd 
be so many separate fields in ``PrivateData`` that it'd be quite messy to look
through. 

> >  I wish I could impose hard limits according to the protocol, but there
> >  aren't clear specifications regarding what is allowed and documentation
> >  for various display models suggest differing limits for display dimensions,
> >  at least according to argument limits for the cursor-shift ``ESC / POS``
> >  commands documented - I don't want to shift the cursor, if requested,
> >  beyond what is allowed by the specification.
> 
> Hm, is there something bad going to happen?

Yes. I've experienced my display crashing on invalid commands
and somehow stop its internal multiplexing VFD drive, driving the last 
5x7 cell continuously on, causing it to be abnormally bright.

These displays are meant to be multiplexed, so I believe this failure
mode will actually degrade the life of the display. Maybe a warning in 
the documentation will suffice, instead of terminating ``LCDproc`` 
during its initialization - I just don't want anyone to lose their pretty 
VFD display.

Best, 

Shenghao Yang

haraldg:
Shenghao Yang writes:

> > I don't think there is anything wrong with having some features not supported
> > by some protocols. Unless each protocol needs a vastly different state,
> > I think it is fine to have state variables for all possible features in
> > the main private data. - Especially since all protocols are statically linked
> > into the main driver anyway.
> 
> If we're going in this direction then I'd like to define a few more structures
> and include those in the ``PrivateData`` structure for the driver - else there'd 
> be so many separate fields in ``PrivateData`` that it'd be quite messy to look
> through. 

My comments in this regard are just suggestions, feel free to ignore them,
if they turn out to be a bad idea and make the code more complex instead
of less.

So far I have mostly seen state variables for cursor position, brightness,
etc which looked pretty much reuseable between all protocols that support
these features - but admittedly I haven't looked too deeply yet. In the end
it's your code. If you think custom structures for each protocol are
necessary, feel free to use them. Maybe combine them into one union?
 
> > >  I wish I could impose hard limits according to the protocol, but there
> > >  aren't clear specifications regarding what is allowed and documentation
> > >  for various display models suggest differing limits for display dimensions,
> > >  at least according to argument limits for the cursor-shift ``ESC / POS``
> > >  commands documented - I don't want to shift the cursor, if requested,
> > >  beyond what is allowed by the specification.
> > 
> > Hm, is there something bad going to happen?
> 
> Yes. I've experienced my display crashing on invalid commands
> and somehow stop its internal multiplexing VFD drive, driving the last 
> 5x7 cell continuously on, causing it to be abnormally bright.

I see. That's more serious then I expected.
 
> These displays are meant to be multiplexed, so I believe this failure
> mode will actually degrade the life of the display. Maybe a warning in 
> the documentation will suffice, instead of terminating ``LCDproc`` 
> during its initialization - I just don't want anyone to lose their pretty 
> VFD display.

Yes. In that case safe default values seem to be the most important.
I will defer to your judgement. But if you decide to leave the limit in the
code, please add a comment explaining this rationale. 

Thanks,
Harald

-- 
If you want to support my work:
see http://friends.ccbib.org/harald/supporting/
or donate via CLAM to xASPBtezLNqj4cUe8MT5nZjthRSEjrRQXN
or via peercoin to P98LRdhit3gZbHDBe7ta5jtXrMJUms4p7w

haraldg:
It seems I didn't get a notification for the updates in this PR. Is this ready for review or do you intend to push more changes?

shenghaoyang:
@haraldg 
I think it's fine, though there may be a few irritating whitespace changes that my poorly configured 
editor sneaked in along with the functionality change commits - sorry about that - it doesn't like the half tab from the switch cases, and GNU indent, which was used to create those half tab stops, likes to 
alter the pointer declarations, but I could fix those up if there's too many.

If you are reviewing them, could you give the docbook files a look through? I copied the syntax
from the original file, but I'm having some trouble building the documentation - ``xmlto`` complains
about being unable to fetch files from sourceforge.

haraldg:
Shenghao Yang writes:
> I think it's fine,

Okay. Then have my apology for not checking up on this for two weeks.

> though there may be a few irritating whitespace changes that my poorly
> configured editor sneaked in along with the functionality change commits
> - sorry about that - it doesn't like the half tab from the switch cases,
> and GNU indent, which was used to create those half tab stops, likes to 
> alter the pointer declarations, but I could fix those up if there's too many.

Don't worry about that. I see the coding style more as guidelines than
a strict law. (Personally I don't bother with half tabs at all, for
example - just have case labels at the same level as the switch statement.)
I think it's enough if people are aware of the coding style and make come
reasonably close.
 
> If you are reviewing them, could you give the docbook files a look through?

Thanks for mentioning this. I'll regenerate the documentation locally, before
merging, then.

> I copied the syntax from the original file, but I'm having some trouble
> building the documentation - ``xmlto`` complains
> about being unable to fetch files from sourceforge.

That's strange. I don't remember this happening to me.

-- 
If you want to support my work:
see http://friends.ccbib.org/harald/supporting/
or donate via CLAM to xASPBtezLNqj4cUe8MT5nZjthRSEjrRQXN
or via peercoin to P98LRdhit3gZbHDBe7ta5jtXrMJUms4p7w

shenghaoyang:
> Okay. Then have my apology for not checking up on this for two weeks.

No problem. Thanks for taking the time to look through the new commits.

> That's strange. I don't remember this happening to me.

Hmm. Maybe I followed the wrong instructions? There was a file in the docs directory that suggested running xmlto with some options to generate a single HTML file.

haraldg:
Shenghao Yang writes:
> Hmm. Maybe I followed the wrong instructions? There was a file in the docs
> directory that suggested running xmlto with some options to.generate a 
> single HTML file.

I see. I usually generate the documentation with "make install-html-guides" ...

-- 
If you want to support my work:
see http://friends.ccbib.org/harald/supporting/
or donate via CLAM to xASPBtezLNqj4cUe8MT5nZjthRSEjrRQXN
or via peercoin to P98LRdhit3gZbHDBe7ta5jtXrMJUms4p7w

haraldg:
This was a bit tedious to review, but I think it is reasonable and will
merge it as is.

Just as feedback for the future: Some of your code is a bit complicated
from either what I feel are unnecessary levels of abstraction or a bit
roundabout programming idioms. The worst example (for the latter), that
I found, was:
+       /* Flush brightness data*/
+       int target_brightness = 0x20;
+       if (data->buffered_misc_state.backlight_state != BACKLIGHT_OFF) {
+               switch (serialPOS_convert_brightness(
+                   data->buffered_misc_state.brightness)) {
+                   /*
+                    * Not the best mapping, but it'll have to do, I don't
+                    * want to get SIGFPE or something
+                    */
+                   case 0x01:
+                       target_brightness = 0x20;
+                       break;
+                   case 0x02:
+                       target_brightness = 0x40;
+                       break;
+                   case 0x03:
+                       target_brightness = 0x60;
+                       break;
+                   case 0x04:
+                       target_brightness = 0xff;
+                       break;

This is more naturally handled by a lookup table. (Also compilers can't
quite optimize it good enough, so this causes an unnecessary memory cost
in the compiled binaries.) I decided I don't care in the case of a
(rather exotic) driver, but I probably wouldn't merge this if it was to
go into the server core.

Anyway: Thanks for you work - this is a huge improvement!

Harald

shenghaoyang:
Thanks for letting me know about that one. Didn't know why I didn't stick
one in, but I'll add a commit and convert that into a lookup.

I don't mind changing anything else that's unnecessarily complicated, as
I've quite a bit of free time, so if there's anything else glaring, please
let me know.

haraldg:
Shenghao Yang writes:
> Thanks for letting me know about that one. Didn't know why I didn't
> stick one in, but I'll add a commit and convert that into a lookup.

Maybe you also want to add yourself to the CREDITS.md file...

> I don't mind changing anything else that's unnecessarily complicated,
> as I've quite a bit of free time, so if there's anything else glaring,
> please let me know.

I always feel unsure how much control to enforce during reviews. On
one hand I feel reviewing should be mentoring and I should help people
improve their programming technique. On the other hand everybody has
their own way of thinking and what might look like a natural solution
to me, might seem roundabout to somebody else. Since I hope that people
will maintain their code in the long run, I don't want to force some
programming style on them that might seem alien to them.

Anyway: In this case your code is one of the better parts of what we
have in lcdproc, so I guess there is little point in spending effort on
cleaning it up even more. If you have some time to spend, I'd rather
prefer if you try to fix some real issues or clean up some of the
really ugly code. (Even if this means you work on some other FLOSS
project than lcdproc.)

Harald

PR #102, opened 2018-07-29
Add option to display or not seconds in big clock

coliss86:
I test this PR on a HD44780 20x4 display. 

If seconds are not displayed, the clock is centered (line 404).


haraldg:
First of all: Documentation is missing. You need to describe what
this configuration setting does in the manual page and the user
guide.

Secondly: I can't review this because actually it's not clear
to me, how this is intended to work. You only want a way to
disable the display of seconds on big displays?

Thanks for working on lcdproc.

Harald

coliss86:
> First of all: Documentation is missing. You need to describe what this configuration setting does in the manual page and the user guide.

I didn't add more documentation because I can't find where the `lcdproc.conf` is described in the user or dev documentation, it seems there is only an example in `clients/lcdproc/lcdproc.conf`.

> Secondly: I can't review this because actually it's not clear to me, how this is intended to work. You only want a way to disable the display of seconds on big displays?

Yes, this is correct. I only want to disable the displays of seconds when the option  `showSecs` is set to `true` in the `lcdproc.conf`.

Here is two photos of the display : 
 * default behavior


* with the option `showSecs` set to `true`

haraldg:
On 30.07.2018 18:33, coliss86 wrote:
> I didn't add more documentation because I can't find where the
> `lcdproc.conf` is described in the user or dev documentation, it 
> seems
> there is only an example in `clients/lcdproc/lcdproc.conf`.

There really seems to be little information in the user guide about the
client. I'm surprised, but whatever.

I still think documentation should be added at least to 
docs/lcdproc.1.in -
yes this file is a bit of a mess at the moment, but that's no excuse to
keep sloppy practises ... ;)

Also your comment in the example file is a bit vague. It should be 
clear
what the setting does without needing to read the code.

>> You only want a way to disable the display of seconds on big 
>> displays?
>
> Yes, this is correct.

Okay.

About your code:

The combination
+       static int showSecs = 1;
+       showSecs = config_get_bool("BigClock", "showSecs", 0, 1);

doesn't work (or rather doesn't make sense). It would reinitialize the
variable on every function call as if it wasn't static at all. I think,
you want to merge these lines together. Ie initialize the variable 
directly
with a call to config_get_bool().

coliss86:
@haraldg what do you think of thoses modifications ? Especially the documentation added to `lcdproc-config.5.in` ? I took the example file, rework it to the man format and added the screen example from sources.

PR #103, opened 2018-08-17
HD44780 FTDI support for USB description and serial number filtering.

jmptbl:
New string configuration directive "UsbDescription" will filter for
the first USB device that has a matching USB description.  Config
"SerialNumber" same convention as from BWCT driver.  Both unconfigured
retains same behaviour as before.

We use FTDI interfaces for a lot of hardware projects, and without this patch it is very tricky to get LCDd to open the correct one!

### DIFF
--- a/docs/lcdproc-user/drivers/hd44780.docbook
+++ b/docs/lcdproc-user/drivers/hd44780.docbook
@@ -4042,13 +4042,25 @@ This can be done by specifying <option>--enable-drivers=all</option> or by inclu
 
 <varlistentry>
   <term>
+    <property>UsbDescription</property> =
+    <parameter><replaceable>DESCRIPTION</replaceable></parameter>
+  </term>
+  <listitem><para>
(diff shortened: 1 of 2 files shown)


haraldg:
I think this is a good idea. Thanks for looking into this!

I have thought if this somehow could be added to all libftdi based drivers,
but I guess it is not all that simple without writing lots of abstraction
code, so didn't follow the idea further.

Your code can be simplified by either using strdup() for getting rid
of strlen()+malloc()+strcpy(), or alloca() for getting rid of this
complicated error handling.

I'd recommend using strdup() but am fine with either way. However I think
one of these optimizations should be used. Other then that, this PR
looks good to me.

jmptbl:
Yes, strdup is much cleaner, thanks! How does it look now?

haraldg:
Looks good, I have merged this now. Thanks for working on this.

PR #106, opened 2018-09-07
Added option to omit hostname in title bar

mobrembski:
I have 20x4 display and i don't like scrolling title bar in each screen of lcdproc just to show my hostname.
So i propose this patch to add ability to control user whenever he want's hostname in title or not.

### DIFF
--- a/clients/lcdproc/main.c
+++ b/clients/lcdproc/main.c
@@ -126,12 +126,14 @@ char *configfile = NULL;
 char *pidfile = NULL;
 int pidfile_written = FALSE;
 char *displayname = NULL;	/**< display name for the main menu */
+int showhostname = TRUE;
+char *hostname = NULL;
 
 /** Returns the network name of this machine */
 const char *
 get_hostname(void)
 {
-	return (unamebuf.nodename);
+	return hostname;
 }
 
 /** Returns the name of the client's OS */
@@ -217,7 +219,7 @@ main(int argc, char **argv)
 	opterr = 0;
 
 	/* get options from command line */
-	while ((c = getopt(argc, argv, "s:p:e:c:fhv")) > 0) {
+	while ((c = getopt(argc, argv, "s:p:e:c:fkhv")) > 0) {
 		char *end;
 
 		switch (c) {
@@ -269,6 +271,16 @@ main(int argc, char **argv)
 	if (cfgresult < 0) {
 		fprintf(stderr, "Error reading config file\n");
 		exit(EXIT_FAILURE);
+	}
+	if (showhostname == TRUE)
+	{
+		hostname = malloc(strlen(unamebuf.nodename)+2);
+		hostname[0] = ' ';
+		strcpy(hostname+1,unamebuf.nodename);
+	}
+	else
+	{
+		hostname = strdup("");
 	}
 
 	/* Set default reporting options */
@@ -422,9 +434,12 @@ process_configfile(char *configfile)
 	if (islow < 0) {
 		islow = config_get_int(progname, "Delay", 0, -1);
 	}
-
-	if ((tmp = config_get_string(progname, "DisplayName", 0, NULL)) != NULL)
+	if ((tmp = config_get_string(progname, "DisplayName", 0, NULL)) != NULL) {
 		displayname = strdup(tmp);
+	}
+	if (showhostname != FALSE) {
+		showhostname = config_get_bool(progname, "ShowHostname", 0, TRUE);
+	}
 
 	/*
 	 * check for config file variables to override all the sequence
@@ -499,6 +514,7 @@ exit_program(int val)
 	 */
 	eyebox_clear();
 #endif
+	free(hostname);
 	Quit = 1;
 	sock_close(sock);
 	mode_close();
--- a/clients/lcdproc/mem.c
+++ b/clients/lcdproc/mem.c
@@ -117,8 +117,11 @@ mem_screen(int rep, int display, int *flags_ptr)
 
 		/* flip the title back and forth... (every 4 updates) */
 		if (which_title & 4) {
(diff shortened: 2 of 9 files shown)


clients/lcdproc/main.c
```diff
@@ -248,6 +250,9 @@ main(int argc, char **argv)
 			case 'f':
 				foreground = TRUE;
 				break;
+			case 'k':
+				showhostname = FALSE;
+				break;
```
haraldg: If you add a command line option, it would need to be documented in the man page.

However unless there is a specific reason why you need a command line option, I recommend to only control this setting from the configuration file.

clients/lcdproc/main.c
```diff
@@ -423,8 +438,9 @@ process_configfile(char *configfile)
 		islow = config_get_int(progname, "Delay", 0, -1);
 	}
 
-	if ((tmp = config_get_string(progname, "DisplayName", 0, NULL)) != NULL)
-		displayname = strdup(tmp);
```
haraldg, line 427: What about this? Why is this code removed?

clients/lcdproc/main.c
```diff
@@ -499,6 +516,7 @@ exit_program(int val)
 	 */
 	eyebox_clear();
 #endif
+	free(hostname);
```
haraldg, line 517: I wouldn't bother with freeing here. If you don't call free, you also don't need to strdup() the empty string above...

clients/lcdproc/mem.c
```diff
@@ -117,8 +117,11 @@ mem_screen(int rep, int display, int *flags_ptr)
 
 		/* flip the title back and forth... (every 4 updates) */
 		if (which_title & 4) {
-			sock_printf(sock, "widget_set M subtitle %i 1 {}\n", lcd_wid - 7);
-			sock_printf(sock, "widget_set M title {%s}\n", get_hostname());
+			if (strcmp(get_hostname(), "") != 0)
+			{
```
haraldg, line 121: Just check for get_hostname()[0] == 0, this is much simpler.

Also I wouldn't object, if you made hostname accessible everwhere and got rid of get_hostname() at all.

haraldg (changes requested):
Seems to be a good idea for everybody using lcdproc on only one machine.  Thanks for looking into this!

A couple of comments, mostly minor nits, inline.

clients/lcdproc/main.c
mobrembski: Ok :) I will remove it

clients/lcdproc/main.c
mobrembski: Sorry, my bad. Added again

clients/lcdproc/mem.c
mobrembski: Fixed. But i think sticking with get_hostname() is more nicer solution for me, i don't like shared variables between units.

clients/lcdproc/main.c
mobrembski: But it is much nicer and elegant to clear used resources before exiting program. I know it doesn't change because program is beeing closed in a moment, but it is more professional.

This is also i've introduced another pull request for fixing detected leaks with valgrind.

mobrembski:
Updated commit

clients/lcdproc/main.c
```diff
@@ -126,12 +126,14 @@ char *configfile = NULL;
 char *pidfile = NULL;
 int pidfile_written = FALSE;
 char *displayname = NULL;	/**< display name for the main menu */
+int showhostname = TRUE;
```
haraldg, line 129: Why do you need to initialze showhostname?

clients/lcdproc/main.c
```diff
@@ -217,7 +219,7 @@ main(int argc, char **argv)
 	opterr = 0;
 
 	/* get options from command line */
-	while ((c = getopt(argc, argv, "s:p:e:c:fhv")) > 0) {
+	while ((c = getopt(argc, argv, "s:p:e:c:fkhv")) > 0) {
```
haraldg, line 222: I think this is a left-over from the deleted code?

clients/lcdproc/main.c
```diff
@@ -270,6 +272,16 @@ main(int argc, char **argv)
 		fprintf(stderr, "Error reading config file\n");
 		exit(EXIT_FAILURE);
 	}
+	if (showhostname == TRUE)
+	{
+		hostname = malloc(strlen(unamebuf.nodename)+2);
+		hostname[0] = ' ';
+		strcpy(hostname+1,unamebuf.nodename);
```
haraldg, line 279: Please keep to the coding style. Ie add a few spaces in the right places.

clients/lcdproc/main.c
```diff
@@ -422,9 +434,12 @@ process_configfile(char *configfile)
 	if (islow < 0) {
 		islow = config_get_int(progname, "Delay", 0, -1);
 	}
-
-	if ((tmp = config_get_string(progname, "DisplayName", 0, NULL)) != NULL)
+	if ((tmp = config_get_string(progname, "DisplayName", 0, NULL)) != NULL) {
 		displayname = strdup(tmp);
+	}
+	if (showhostname != FALSE) {
```
haraldg, line 440: I think this if statement is not needed after removing the command line option.

clients/lcdproc/main.c
haraldg: Well, C is often not nice or elegant, but practical ... :-)

I disagree about professionalism: I'm fine with adding extra code, if it helps readability (though comments are usually better), but in this case it only inflates the source and the footprint of the binary without any true benefit.

haraldg (commented):
Looks like we need an other round of updates.

PR #107, opened 2018-09-07
Fix lcdproc memory leaks and warnings

mobrembski:
Fixes memory leaks and compilation warnings in lcdproc

### DIFF
--- a/clients/lcdproc/iface.c
+++ b/clients/lcdproc/iface.c
@@ -517,3 +517,11 @@ actualize_transfer_screen(IfaceInfo *iface, int index)
 		}
 	}
 }
+
+void iface_shutdown(void)
+{
+	for (iface_count = 0; iface_count < MAX_INTERFACES; iface_count++) {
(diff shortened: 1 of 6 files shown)


haraldg:
Hi,

thanks for working on lcdproc. It's not clear to me, what problems this
patch is trying to solve. (Note that it is not necessary to free() memory
immediatly before terminating the process, as terminating frees all RAM
(and file descriptors) anyway.

It seems there are different independent changes in this patch. Maybe
you can submit the real fixes as separate patches - can be one PR or
even overwritting this PR, but different commits - with commit messages
describing what actual problem is solved by the commit.

Thanks,
Harald

willfe:
I'm closing this for inactivity (it's been 8 years) but have generated a new task for myself (#229) to go digging through the repo for memory leaks and compiler warnings.

Thanks OP for submitting this (and @haraldg for responding to it back in the day). I'll try to get things moving again to hopefully address things you found when you submitted this to the project.

PR #111, opened 2018-10-05
Added an option to turn on/off the backlight for the oldtime display.

NicolasJourden:


clients/lcdproc/lcdproc.conf
```diff
@@ -106,7 +106,8 @@ DateFormat="%x"
 # Display the title bar in two-line mode. Note that with four lines or more
 # the title is always shown. [default: true; legal: true, false]
 #ShowTitle=false
-
+# Enable or disable the backlight
```
dod38fr: Could you add `[default: false; legal: true, false]` in this comment ?

This helps [cme](https://github.com/dod38fr/config-model/wiki/Managing-Lcdproc-configuration-with-cme) to [generate a more accurate model for lcdproc](https://ddumont.wordpress.com/2011/07/03/generate-a-configuration-editor-from-a-config-template-file-with-perl-lcdproc-example/)

All the best

haraldg:
What's the rationale behind this change? Ie why would you want to set the backlight only for the oldtime screen but not the others?

clients/lcdproc/lcdproc.conf
NicolasJourden: This has been added in the second commit.

NicolasJourden:
> What's the rationale behind this change? Ie why would you want to set the backlight only for the oldtime screen but not the others?

Sure. I could add this option to all the screens related in chrono.c
What would you suggest? Or add this to all the screens?
To me, by default the backlight should be on.
And you can set it off using backlight parameter in each of the screens of lcdproc.conf.

I added this option because I use the LCD as a display for my alarm clock.
And I have another screen made using telnet that turns on the screen and display nice things.

haraldg:
Nicolas JOURDEN writes:
> Sure. I could add this option to all the screens related in chrono.c
> What would you suggest? Or add this to all the screens?
> To me, by default the backlight should be on.
> And you can set it off using backlight parameter in each of the screens of lcdproc.conf.

Well usually people set the backlight (default) in their LCDd.conf and then
they can change it from the menu at runtime (if they have keys).

Clients can change the backlight on a per screen basis but AFAIK it is
rarely done. Do you actually have a use case where you want to have one
lcdproc screen with backlight on and an other lcdproc screen with backlight
off?

> And I have another screen made using telnet that turns on the screen and display nice things.

Anything you want to share? We might add your script to the examples
folder ...

NicolasJourden:
Yes, I can manage to share my perl script showing how I configure with second screen.

My use case is simple: display the hour, date on the screen with backlight off by default.
Run another client that display another information (title played) with the backlight on.

Another solution for that could be hardware too since I could control the backlight with a simple PWM and a transistor.

Also I noticed that clients can't control each other screens. Meaning my client can not control the backlight for OldTime.

haraldg:
Nicolas JOURDEN writes:
> My use case is simple: display the hour, date on the screen with backlight off by default.
> Run another client that display another information (title played) with the backlight on.

Then I think adding a backlight option on a per client basis should be enough ...

> Also I noticed that clients can't control each other screens. Meaning my client can not control the backlight for OldTime.

Yes, it's debateable whether the client isolation in lcdproc is a good
thing. We already have "enable_permissive_goto" to work around that. I'm
open to ideas to poke an other hole into client isolation, if there is
interest in doing so.

haraldg:
Never anything came from this discussion so closing. Please open an new PR if there is any code implementing one of the above ideas. 

If you feel there is a need for further discussion, feel free to reopen this PR.

PR #116, opened 2019-04-11
Small screen and mx5000 fixes

jwrdegoede:
Here is a series of fixes for screens with 4+ lines, while being smaller then 18 characters wide as well as a some fixes + input support for the mx5000 driver.

This series has been tested with a mx5000 keyboard which has a builtin LCD with a size of 17x4 characters.

### DIFF
--- a/clients/lcdproc/cpu.c
+++ b/clients/lcdproc/cpu.c
@@ -75,9 +75,9 @@ cpu_screen(int rep, int display, int *flags_ptr)
 			sock_send_string(sock, "widget_add C two string\n");
 			sock_send_string(sock, "widget_add C three string\n");
 			sock_printf(sock, "widget_set C one 1 2 {%-*s%-*s}\n",
-					lcd_wid / 2, "Usr", lcd_wid / 2, "Nice");
+					(lcd_wid + 1) / 2, "Usr", lcd_wid / 2, "Nice");
 			sock_printf(sock, "widget_set C two 1 3 {%-*s%-*s}\n",
-					lcd_wid / 2, "Sys", lcd_wid / 2, "Idle");
+					(lcd_wid + 1) / 2, "Sys", lcd_wid / 2, "Idle");
 			sock_printf(sock, "widget_set C three 1 4 {0%%%*s100%%}\n", gauge_wid, "");
 			sock_send_string(sock, "widget_add C usr string\n");
 			sock_send_string(sock, "widget_add C nice string\n");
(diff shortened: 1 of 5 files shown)


jwrdegoede:
FYI I've also submitted a set of fixes to libmx5000, this series is independent of those fixes:
https://github.com/coolacid/mx5000tools-revo/pull/1

haraldg:
Looks very good. I have only one point of concern: The "Add support for buttons ..." patch makes this explicitely a linux only driver. OTOH we already have a generic linux_input driver, which might be used as well.

I guess it is no big deal in this case, as mx5000tools/libmx5000 seems to be linux only itself at the moment. Still it feels cleaner to pretend this driver is generic.

What do you think?

jwrdegoede:
I realize that the same functionality can probably (I did not try) be achieved with the linux_input driver. I did actually notice that driver before I wrote the "Add support for buttons ..." patch. But that will require a lot of manual setup and I prefer an approach which will "just work" for end users.

Since I got my mx5000 keyboard I've actually been working on some auto-configuration and activation of LCDd and lcdproc, so that we can **maybe** ship them as part of the default Fedora workstation live images and have LCD status screens "just work" ootb in some cases. Here is what I have sofar: 
https://fedorapeople.org/~jwrdegoede/lcdproc/

The idea is that this work as follows, the LCDd*.service and lcdproc.service will be installed in /lib/systemd/system but they will not get enabled by default (users can still enable them manually when using a manual config).

90-lcdproc.rules is a udev rule which will get lcdproc.target started explicitly which will start LCDd-hwdetect.service LCDd.service and lcdproc.service in that order. LCDd-hwdetect.service will generate a LCDd.conf *if none exists* using LCDd-hwdetect.service, which knows about the same LCD-s as the udev rules. I do plan to upstream this eventually but atm it still is very new and only supports 1 device. I do have a 
a Logitech g15 v1 keyboard and one of these: https://www.banggood.com/1602-RGB-LCD-Display-With-USB-Port-For-Raspberry-Pi-3B-2B-B-Windows-Linux-p-1079617.html on the way and I plan to also add auto detect/config of those too.

Having input support inside the mx5000 driver which "just works" will be quite helpful for this and IMHO the knowledge of the specific custom Logitech scancodes used for the buttons below the LCD really belongs inside the mx5000 driver.

Given that the new input support is limited to just 2 functions, if libmx5000 ever gets supported on other platforms we can easily wrap these 2 in #ifdef __linux__ .

FYI this whole adventure started because I gota mx5000 keyboard to have a Logitech bluetooth keyboard for testing and developing a large kernel patch series which I've been working on which greatly improves support for Logitech wireless keyboards and mice, see: https://www.spinics.net/lists/linux-input/msg60944.html

haraldg:
Since you clearly have considered linux_input, but still prefer to handle
this in the mx5000 driver, that's good enough for me. I intend to merge
this PR.

However now that I know the big picture of what you are doing (thanks
BTW) an other question comes to mind: Wouldn't it be better to handle
this in LCDd-hwdetect.service? lcdproc releases way slower than the
kernel and probably slower than many distributions. If new HW revisions
use different scan codes or some other minor incompatibility happens, I'd
figure most distributions would prefer patching LCDd-hwdetect.service
instead of the driver itself.

jwrdegoede:
I really want to limit the CDd-hwdetect.sh script to mapping USB/Bluetooth product and device ids to drivers.

Also the current linux_input driver does not support mappings based on raw scancodes at all, and by default the kernel will send KEY_UNKNOWN as event-code for all 4 keys below the mx5000 LCD so we really need to look at the raw scancodes. As for the scancodes changing with different revisions I'm not worried about that.

So all in all I would like to proceed with this pull-req as is.

haraldg:
This still seems a bit hacky, but I have merged it anyway on the grounds that it is unclear which way future development (like BSD support) would be going. This might be subject to refactoring in the future.

jwrdegoede:
Thank you for merging this.

Issue #74, opened 2017-03-26
suggested changes to the protocol (client ABI)


haraldg:
This ticket is to collect all the ideas for changes to the network protocol (backwards compatible or not). Also maybe to collect current and future real world use cases.

# Character encoding (character set?)
Currently LCDd just shuffles bytes between clients and drivers and it is the responsibility of the user to set up everything, so that things fit together. This means we are mostly using 7bit ascii, but exceptions exist. There should be a way to at least negotiate what's supported. Even better if LCDd could be told to do some conversions.

# Frames
are specified in the current version of the protocol, but most parts never got implemented (for more then a decade). Maybe it is time to acknoledge that we target small displays where these are not very useful and nobody really needs them. We should figure out which parts are actually implemented (and work!) and drop the rest from the ABI and also remove any code that never actually worked as intended.

# Way to edit times (or strings)
This could be a new menu widget or a general input widget (lcdproc currently doesn't have general input widgets though). Either it would be special for editing time strings like we already have for IPv4 addresses or it would take a specification of allowed characters in the string to be edited. The latter would allow to not only edit times, but also cron lines, etc.

Example for simple specification syntax: `[012][0-9]:[0-5][0-9]`  
This example has the problem that "29:00" is not a valid time. The purpose of the specification would be more to make the editing process easier with only Up/Down keys easier, than to ensure that only legal values can be entered. A more complex specification language would be necessary for the latter. Not sure if it is worth the effort, though.

haraldg:
An other issue with the current protocol, that makes me slightly uncomfortable, is the way how key reservation works:
* `client_add_key` takes an arbitrary number of keys (strings) as argument, returns a "huh?" message for each key that it can't grab and finally an unconditional "success". "huh?" messages aren't meant for machine parsing, so the client doesn't actually know which keys it now owns.
* If the client requested just one key, we return a "huh?" message and "success", which is unambiguous but bad taste IMO.
* keys are requested on the level of clients not screens. Maybe instead of the -shared/-exclusively distinction there should be a screen/client level distinction: Shared keys would be requested per screen and only returned when this very screen is visible.

emteejay:
Graphic displays also need consideration. It should be possible for positions and sizes to be specified in pixels rather than characters and lines, also scrolling should be in pixel steps.

haraldg:
Martin Jones writes:
[...]

PR #117, opened 2019-04-17
Bar widgets rework

jwrdegoede:
This pull-req extends the "widget_set screen widget_name parm1 parm2 parm3" syntax for hbars and vbars with an optional new  "widget_set screen widget_name parm1 parm2 parm3 parm4" syntax.

The reason for this is to address the current widget_set syntax for bar widgets and the driver vbar / hbar being mismatched wrt each other and to avoid the hacks / problems this mismatch causes.

Specifically the old 3 parameter widget_set syntax for bars takes the usual x and y parameters as parm1 and parm2 and then the 3th argument is the length of the filled-part of the bar to draw in pixels. The drivers however expect x, y, length of the entire bar (not just the filled part) in characters and a filled promillage.

To workaround this mismatch, the render code in the server currently has this hack for hbars:
```
		int full_len = display_props->width - w->x - left + 1;
		int promille = 1000;

		if ((w->length / display_props->cellwidth) < right - left - w->x + 1)
			promille = (long) 1000 * w->length / (display_props->cellwidth * full_len);

		drivers_hbar(w->x + left, w->y + top, full_len, promille, BAR_PATTERN_FILLED);
```

This tells the driver to render a hbar with a length which reaches all the way to the right edge of the display. If you look at the 4 line standard cpu screen, this means that the length overwrites the "100%" text after the bar on the 4th line. Likewise for the 4 line standard mem screen, which has 2 small bars on the bottom, the mem bar overwrites the F E [swap-hbar ] F part of the display and the swap-bar length overwrites the F at the end of the 4th line. This all works without the 100% / F E [] F text which is only rendered once disappearing because most drivers only render the full part of the hbar, simply never drawing the unfilled-part of the bar, relying on that part of the LCD being empty. Except it is not empty, but not overwriting what is there is desirable because of the widget_set vs driver API mismatch.

So this mostly works as is, but IMHO the current state is a big hack. And this hack is causing problems for some drivers, specifically for the mx5000 driver. libmx5000 has its own hbar implementation and the lcdproc mx5000 driver is just a wrapper around this. But the mx5000 hbars consist of special begin (full or empty) middle (full or empty) and end characters, see:

https://fedorapeople.org/~jwrdegoede/cpu.png
https://fedorapeople.org/~jwrdegoede/mem.png

Note these pictures are how things look **with** the patches from this pull-req. The mx5000 code actually renders empty middle and end chars for the non-filled part of the hbar, breaking the assumption of the code working around the API mismatch that for the non-filled part nothing is rendered. This means that the end of the hbar overwrites the 100% in the CPU screen and the F E [] F text in the mem screen also gets overwritten which looks really ugly.

The first commit in this series fixes the API mismatch by introducing a new 4 parameter widget_set syntax for bars which matches the driver API. This new  widget_set syntax for bars is IMHO also easier to use by various screen code since it avoids the need to convert something like a CPU load to a filled-part of the bar in pixels unit.

The other commits in this pull-req convert the hbar usage in the standard lcdproc client cpu and mem screens to use the new syntax, fixing the issues with those screens on the mx5000.

Note that support for the old 3 parameter syntax is kept unchanged, to avoid clients relying on that breaking.

Besides fixing the mx5000 hbars this will also allow other (graphical) displays to potentially draw nicer bars then currently is done.

### DIFF
--- a/server/commands/widget_commands.c
+++ b/server/commands/widget_commands.c
@@ -296,6 +296,31 @@ widget_set_func(Client *c, int argc, char **argv)
 			sock_send_string(c->sock, "success\n");
 		}
 		break;
+	case WID_PBAR:			/* Pbar takes "x y width promille [begin-label end-label]" */
+		if (argc < i + 4 || argc > i + 6) {
+			sock_send_error(c->sock, "Wrong number of arguments\n");
+			break;
+		}
+		if ((!isdigit((unsigned int) argv[i][0])) ||
+		    (!isdigit((unsigned int) argv[i + 1][0]))) {
+			sock_send_error(c->sock, "Invalid coordinates\n");
+			break;
+		}
+		free(w->begin_label);
+		free(w->end_label);
+		w->begin_label = NULL;
+		w->end_label = NULL;
+		w->x = atoi(argv[i]);
+		w->y = atoi(argv[i + 1]);
+		w->width = atoi(argv[i + 2]);
+		w->promille = atoi(argv[i + 3]);
+		if (argc >= i + 5)
+			w->begin_label = strdup(argv[i + 4]);
+		if (argc >= i + 6)
+			w->end_label = strdup(argv[i + 5]);
+		debug(RPT_DEBUG, "Widget %s set to %i", wid, w->promille);
+		sock_send_string(c->sock, "success\n");
+		break;
 	case WID_ICON:			/* Icon takes "x y icon" */
 		if (argc != i + 3)
 			sock_send_error(c->sock, "Wrong number of arguments\n");
--- a/server/driver.c
+++ b/server/driver.c
@@ -56,6 +56,7 @@ DriverSymbols driver_symbols[] = {
 	{ "chr",                offsetof(Driver, chr),                0 },
 	{ "vbar",               offsetof(Driver, vbar),               0 },
 	{ "hbar",               offsetof(Driver, hbar),               0 },
+	{ "pbar",               offsetof(Driver, pbar),               0 },
 	{ "num",                offsetof(Driver, num),                0 },
 	{ "heartbeat",          offsetof(Driver, heartbeat),          0 },
 	{ "icon",               offsetof(Driver, icon),               0 },
@@ -447,6 +448,63 @@ driver_alt_hbar(Driver *drv, int x, int y, int len, int promille, int options)
 	}
 }
 
+/**
+ * Draw a percentage-bar including its (optional) labels.
+ * \param drv          Pointer to driver structure.
+ * \param x            Horizontal character position (column) of the starting point.
+ * \param y            Vertical character position (row) of the starting point.
+ * \param width        Width of the widget in characters, including the
+ *                     optional begin and end-labels.
+ * \param promille     Current length level of the bar in promille.
+ * \param begin_label  Optional (may be NULL) text to render in front of /
+ *                     at the beginning of the percentage-bar.
+ * \param end_label    Optional text to render at the end of the pbar.
+ */
+void
+driver_pbar(Driver *drv, int x, int y, int width, int promille, char *begin_label, char *end_label)
+{
+	int begin_length, end_length, len;
+
+	debug(RPT_DEBUG, "%s(drv=[%.40s], x=%d, y=%d, width=%d, promille=%d)", __FUNCTION__, drv->name, x, y, width, promille);
+
+	/* if the driver does not support output, do nothing. */
+	if (drv->chr == NULL || drv->string == NULL)
+		return;
+
+	/* if we are falling back to drawing a hbar because the driver does
+	 * not provide a pbar method *and* no labels are given, draw []
+	 * around the hbar to mark the beginning and end of the bar.
+	 */
+	if (!drv->pbar && begin_label == NULL && end_label == NULL) {
+		begin_label = "[";
+		end_label = "]";
+	}
+
+	begin_length = begin_label ? strlen(begin_label) : 0;
+	end_length = end_label ? strlen(end_label) : 0;
+
+	/* We want at least 2 chars for the bar itself */
+	if ((begin_length + end_length + 2) > width)
+		begin_length = end_length = 0;
+
+	len = width - begin_length - end_length;
+
+	if (begin_length) {
+		drv->string(drv, x, y, begin_label);
+		x += begin_length;
+	}
+
+	if (drv->pbar)
+		drv->pbar(drv, x, y, len, promille);
+	else if (drv->hbar)
+		drv->hbar(drv, x, y, len, promille, BAR_PATTERN_FILLED);
+	else
+		driver_alt_hbar(drv, x, y, len, promille, BAR_PATTERN_FILLED);
+	x += len;
+
+	if (end_length)
+		drv->string(drv, x, y, end_label);
+}
 
 /** Write a big number to the screen.
  * Fallback for the driver's \c num method if the driver does not provide one.
--- a/server/drivers.c
+++ b/server/drivers.c
@@ -310,6 +310,30 @@ drivers_hbar(int x, int y, int len, int promille, int pattern)
 
 
 /**
+ * Draw a percentage-bar to all drivers.
+ * \param x            Horizontal character position (column) of the starting point.
+ * \param y            Vertical character position (row) of the starting point.
+ * \param width        Width of the widget in characters, including the
+ *                     optional begin and end-labels.
+ * \param promille     Current length level of the bar in promille.
+ * \param begin_label  Optional (may be NULL) text to render in front of /
+ *                     at the beginning of the percentage-bar.
+ * \param end_label    Optional text to render at the end of the pbar.
+ *
+ * Note the driver may choose to not render the labels if there is not enough
+ * space.
+ */
+void
+drivers_pbar(int x, int y, int width, int promille, char *begin_label, char *end_label)
+{
+	Driver *drv;
+
+	ForAllDrivers(drv)
+		driver_pbar(drv, x, y, width, promille, begin_label, end_label);
+}
+
+
+/**
  * Write a big number to all output drivers.
  * For drivers that define a num() function, call it;
  * otherwise call the general driver_alt_num() function from the server core.
--- a/server/widget.c
+++ b/server/widget.c
@@ -33,6 +33,7 @@ char *typenames[] = {
 	"string",	/* WID_STRING */
 	"hbar",		/* WID_HBAR */
 	"vbar",		/* WID_VBAR */
+	"pbar",		/* WID_PBAR */
 	"icon",		/* WID_ICON */
 	"title",	/* WID_TITLE */
 	"scroller",	/* WID_SCROLLER */
@@ -85,7 +86,7 @@ widget_create(char *id, WidgetType type, Screen *screen)
 	debug(RPT_DEBUG, "%s(id=\"%s\", type=%d, screen=[%s])", __FUNCTION__, id, type, screen->id);
 
 	/* Create it */
-	w = malloc(sizeof(Widget));
+	w = calloc(1, sizeof(Widget));
 	if (w == NULL) {
 		report(RPT_DEBUG, "%s: Error allocating", __FUNCTION__);
 		return NULL;
--- a/configure.ac
+++ b/configure.ac
@@ -246,7 +246,7 @@ AC_ARG_WITH(lcdport,
 )
 AC_DEFINE_UNQUOTED(LCDPORT, $LCDPORT, [Set default port where LCDd should listen])
 
-AC_DEFINE_UNQUOTED(PROTOCOL_VERSION, "0.3", [Define version of lcdproc client-server protocol])
+AC_DEFINE_UNQUOTED(PROTOCOL_VERSION, "0.4", [Define version of lcdproc client-server protocol])
 
 AC_DEFINE_UNQUOTED(API_VERSION, "0.5", [Define version of lcdproc API])
 
--- a/docs/LCDd.8.in
+++ b/docs/LCDd.8.in
@@ -377,7 +377,7 @@ of display available.
 
 .SS CLIENT COMMANDS
 There are many commands for the client to send to the LCDd server, here is
-the list of commands for version \fI0.3\fP of the protocol:
+the list of commands for versions \fI0.3\fP - \fI0.4\fP of the protocol:
 .TP
 .B hello
 This starts a client-server session with the LCDd server; the server will
@@ -389,7 +389,7 @@ The first 4 keyword value pairs are always wid, hgt, cellwid and cellhght,
 for backwards compatibility with older clients which are hardcoded to take
 the 8th and 10th words of the connect string as width and height. The
 connect string will look like this for example:
-"\fBconnect LCDproc 0.5dev protocol 0.3 lcd wid 17 hgt 4 cellwid 6 cellhgt 10\fP"
+"\fBconnect LCDproc 0.5dev protocol 0.4 lcd wid 17 hgt 4 cellwid 6 cellhgt 10\fP"
 .TP
 .B client_set -name \fIname\fP
 Set the client's name.
@@ -508,6 +508,10 @@ A horizontal bar graph.
 .B vbar
 A vertical bar graph.
 .TP
+.B pbar
+A-percentage bar, filled from left to right. This widget-type is only
+available on servers which report a \fIprotocol-version\fP of 0.4 or newer.
+.TP
 .B title
 A title displayed across the top of the display, within a banner.
 .TP
@@ -542,6 +546,13 @@ Displays a horizontal bar starting at position (\fIx\fP,\fIy\fP) that is \fIleng
 .TP
 .B vbar \fIx y length\fP
 Displays a vertical bar starting at position (\fIx\fP,\fIy\fP) that is \fIlength\fP pixels high.
+.TP
+.B pbar \fIx y width promille\fP [\fIbegin-label end-label\fP]
+Displays a percentage-bar starting at position (\fIx\fP,\fIy\fP) covering
+\fIwidth\fP character cells including the optional begin and end-labels,
+filled from left to right to the specified \fIpromille\fP value.
+Optionally a \fIbegin-label\fP and \fIend-label\fP can be specified which
+will be drawn in front of and after the percentage-bar.
 .TP
 .B icon \fIx y name\fP
 Displays the icon \fIname\fP at position (\fIx\fP,\fIy\fP).
--- a/docs/LCDd.8.in
+++ b/docs/LCDd.8.in
@@ -376,12 +376,20 @@ of display available.
 .PP
 
 .SS CLIENT COMMANDS
-There are many commands for the client to send to the LCDd server:
+There are many commands for the client to send to the LCDd server, here is
+the list of commands for version \fI0.3\fP of the protocol:
 .TP
 .B hello
-This starts a client-server session with the LCDd server; the
-server will return a data string detailing the type of display
-and its size.
+This starts a client-server session with the LCDd server; the server will
+return a white-space separated data string detailing the type of display
+and its size. The format of this string is:
+"\fBconnect LCDproc\fP <\fILCDd-version\fP> \fBprotocol\fP <\fIprotocol-version\fP> \fBlcd\fP [\fIkeyword value\fP] [\fIkeyword value\fP] ...".
+\fIprotocol-version\fP is always in the form of "\fImajor\fP.\fIminor\fP".
+The first 4 keyword value pairs are always wid, hgt, cellwid and cellhght,
+for backwards compatibility with older clients which are hardcoded to take
+the 8th and 10th words of the connect string as width and height. The
+connect string will look like this for example:
+"\fBconnect LCDproc 0.5dev protocol 0.3 lcd wid 17 hgt 4 cellwid 6 cellhgt 10\fP"
 .TP
 .B client_set -name \fIname\fP
 Set the client's name.
--- a/docs/lcdproc-dev/driver-api.docbook
+++ b/docs/lcdproc-dev/driver-api.docbook
@@ -104,6 +104,9 @@ typedef struct lcd_logical_driver {
 	void (*vbar)		(Driver *drvthis, int x, int y, int len, int promille, int options);
 	void (*hbar)		(Driver *drvthis, int x, int y, int len, int promille, int options);
 
+	// draw a progressbar covering lenght chars at pos (x,y) showing promille% progress
+	void (*pbar)		(Driver *drvthis, int x, int y, int len, int promille);
+
 	// display (big) number num at horizontal position x
 	void (*num)		(Driver *drvthis, int x, int num);
 
@@ -396,9 +399,30 @@ typedef struct MyDriver_private_data {
   </funcprototype>
 </funcsynopsis>
 <para>
-  Draw a horizontal bar at position (<replaceable>x</replaceable>,<replaceable>y</replaceable>)
+  Draw a solid horizontal bar at position (<replaceable>x</replaceable>,<replaceable>y</replaceable>)
   that has maximal length <replaceable>len</replaceable>, where a fraction of
-  (<replaceable>promille</replaceable> / 1000) is filled.
+  (<replaceable>promille</replaceable> / 1000) is filled. The non filled
+  part of the maximum length should not be drawn / touched by this function.
+</para>
+
+<funcsynopsis>
+  <funcprototype>
+	<funcdef>void <function>(*pbar)</function></funcdef>
+	<paramdef>Driver *<parameter>drvthis</parameter></paramdef>
+	<paramdef>int <parameter>x</parameter></paramdef>
+	<paramdef>int <parameter>y</parameter></paramdef>
+	<paramdef>int <parameter>len</parameter></paramdef>
+	<paramdef>int <parameter>promille</parameter></paramdef>
+  </funcprototype>
+</funcsynopsis>
+<para>
+  Draw a progress bar covering <replaceable>len</replaceable> characters at
+  position (<replaceable>x</replaceable>,<replaceable>y</replaceable>),
+  where a fraction of (<replaceable>promille</replaceable> / 1000) is filled.
+  Unlike a hbar, this function must draw a background for the the non-filled
+  part of the bar. For example this function may draw an outline in the form
+  of a (non-filled) rectangle covering the entire length and then fill part
+  of that rectangle to indicate the current progress.
 </para>
 
 <funcsynopsis>
--- a/docs/lcdproc-dev/language.docbook
+++ b/docs/lcdproc-dev/language.docbook
@@ -452,6 +452,14 @@
 		</varlistentry>
 		<varlistentry>
 		  <term>
+		    <literal>pbar</literal>
+		  </term>
+		  <listitem><para>
+		      A percentage / progress bar.
+		    </para></listitem>
+		</varlistentry>
+		<varlistentry>
+		  <term>
 		    <literal>icon</literal>
 		  </term>
 		  <listitem><para>
@@ -564,6 +572,27 @@
 		      vertical (<literal>vbar</literal>) starting at
 		      position (<replaceable>x</replaceable>,<replaceable>y</replaceable>)
 		      that is <replaceable>length</replaceable> pixels wide resp. high.
+		    </para></listitem>
+		</varlistentry>
+		<varlistentry>
+		  <term>
+		    <literal>pbar</literal>
+		  </term>
+		  <listitem>
+		    <cmdsynopsis>
+		      <arg choice="plain"><replaceable>x</replaceable></arg>
+		      <arg choice="plain"><replaceable>y</replaceable></arg>
+		      <arg choice="plain"><replaceable>width</replaceable></arg>
+		      <arg choice="plain"><replaceable>promille</replaceable></arg>
+		      <arg choice="opt"><replaceable>begin-label</replaceable></arg>
+		      <arg choice="opt"><replaceable>end-label</replaceable></arg>
+		    </cmdsynopsis>
+		    <para>
+		      Displays a horizontal (<literal>pbar</literal>) and its
+		      optional (<literal>begin-label</literal>) and (<literal>end-label</literal>)
+		      labels covering <replaceable>width</replaceable> characters starting at
+		      position (<replaceable>x</replaceable>,<replaceable>y</replaceable>)
+		      where a fraction of (<replaceable>promille</replaceable> / 1000) is filled.
 		    </para></listitem>
 		</varlistentry>
 		<varlistentry>
--- a/server/driver.h
+++ b/server/driver.h
@@ -46,6 +46,9 @@ driver_support_multiple(Driver *driver);
 bool
 driver_stay_in_foreground(Driver *driver);
 
+void
+driver_pbar(Driver *drv, int x, int y, int width, int promille, char *begin_label, char *end_label);
+
 
 /* Alternative functions for all extended functions */
 
--- a/server/drivers.h
+++ b/server/drivers.h
@@ -50,6 +50,9 @@ void
 drivers_hbar(int x, int y, int len, int promille, int pattern);
 
 void
+drivers_pbar(int x, int y, int width, int promille, char *begin_label, char *end_label);
+
+void
 drivers_num(int x, int num);
 
 void
--- a/server/drivers/lcd.h
+++ b/server/drivers/lcd.h
@@ -156,6 +156,7 @@ typedef struct lcd_logical_driver {
 	/* extended output functions (optional; core provides alternatives) */
 	void (*vbar)		(struct lcd_logical_driver *drvthis, int x, int y, int len, int promille, int pattern);
 	void (*hbar)		(struct lcd_logical_driver *drvthis, int x, int y, int len, int promille, int pattern);
+	void (*pbar)		(struct lcd_logical_driver *drvthis, int x, int y, int width, int promille);
 	void (*num)		(struct lcd_logical_driver *drvthis, int x, int num);
 	void (*heartbeat)	(struct lcd_logical_driver *drvthis, int state);
 	int (*icon)		(struct lcd_logical_driver *drvthis, int x, int y, int icon);
--- a/server/drivers/mx5000.c
+++ b/server/drivers/mx5000.c
@@ -303,17 +303,16 @@ mx5000_backlight (Driver *drvthis, int on)
 }
 
 /*
- * Draws a horizontal bar to the right.
- */
-MODULE_EXPORT void
-mx5000_hbar (Driver *drvthis, int x, int y, int len, int promille, int options)
-{
-/* x and y are the start position of the bar.
+ * Draws a progressbar.
+ * x and y are the start position of the bar.
  * The bar by default grows in the 'right' direction
  * (other direction not yet implemented).
  * len is the number of characters that the bar is long at 100%
  * promille is the number of promilles (0..1000) that the bar should be filled.
  */
+MODULE_EXPORT void
+mx5000_pbar (Driver *drvthis, int x, int y, int len, int promille)
+{
     PrivateData *p = drvthis->private_data;
     // The position should be given in pixels
     int px, py;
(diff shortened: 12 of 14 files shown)


haraldg:
Well, we are supporting external clients. Thus we see the widget language
as ABI and try to not break it without due consideration. At minimum this
PR would need to update all the relevant docs and probably also bump
the protocol version. There is issue #74, which collects ideas for
updates to the command language.

I'm not a fan of our pixel based command parameters, so I'm quite in
favor of your proposal. But OTOH the behavior you observed is quite
clearly a bug, which we should fix independently of new updates to
the widget language. I think in this case we just need to be smarter
about calculating 'full_len' ...

Also when updating the widget language, we should think about bars, that
include text descriptions.

On a more general note: Most of the rendering code is a huge hack, that
never implemented the less common use cases of the command language.
Breakage like you observed happens all the time with custom clients
or on the less tested drivers, because we have lots of code that
depends on one widget overwritting the other in the right way on the
right time. I think the only realistic way to fix this is to drop some
of the functionality from the widget language and reimplement the rest
with lower overall complexity.

Somewhat unrelated: Personally my plan is to provide liblcdproc, which
abstracts many of the rough edges of the widget language away, handles
multiple protocol versions and provides additional convenience functions
for clients. I have started on a draft, but never got far enough to
work on widgets like bars. Also this work is stalled ATM.

jwrdegoede:
On 17-04-19 23:15, haraldg wrote:
> Well, we are supporting external clients. Thus we see the widget language
> as ABI and try to not break it without due consideration.

Right, notice that so far widget_set commands done on bar widgets
have only been valid when they take 5 parameters (screen + widget-name
+ 3 additional parameters) the first patch in this series allows clients
to use a new syntax with 4 additional parameters. Which older servers
will reject, since they reject all widget_set commands on bar widgets
with a parameter-count != 5.

I completely agree that the ABI should not be changed and for old /
unmodified clients which still pass 3 additional parameters nothing
changes.

The new behavior only happens with widget_set commands on bar widgets
which pass 4 additional parameters. Which no client should be doing
so far since such calls have not been valid commands so far.

> At minimum this
> PR would need to update all the relevant docs

I agree the docs need to be updated, that is something I overlooked
so I guess the command set is documented somewhere?

> and probably also bump
> the protocol version.

If that can be done without breaking existing clients? Then yes that
is probably a good idea so that clients can detect if the new 4
param version is supported.

> There is issue #74, which collects ideas for
> updates to the command language.
> 
> I'm not a fan of our pixel based command parameters, so I'm quite in
> favor of your proposal. But OTOH the behavior you observed is quite
> clearly a bug, which we should fix independently of new updates to
> the widget language. I think in this case we just need to be smarter
> about calculating 'full_len' ...

I've thought about being smarter about calculating 'full_len', but
unless we keep a cache of all other stuff drawn and then try to avoid
drawing over that there is just no way to know the correct full_len,
which is why I've rejected that approach.

I've actually given this all careful consideration as I said above
I did take great care to not break the command language while adding
support for the new way of passing the parameters. I chose this route
because being smarter about calculating 'full_len' is almost impossible
to do.

> Also when updating the widget language, we should think about bars, that
> include text descriptions.

Well that would be a whole new type of bar IMHO and thus outside the
scope of this PR.

haraldg:
Hans de Goede writes:
> I agree the docs need to be updated, that is something I overlooked
> so I guess the command set is documented somewhere?

AFAIR the command language is described in docs/LCDd.8.in and
docs/lcdproc-dev ...

> > and probably also bump
> > the protocol version.
> 
> If that can be done without breaking existing clients? Then yes that
> is probably a good idea so that clients can detect if the new 4
> param version is supported.

Yes, clients should check the protocol version reported by the server
and use the proper set of commands (or give up).

> > There is issue #74, which collects ideas for
> > updates to the command language.
> > 
> > I'm not a fan of our pixel based command parameters, so I'm quite in
> > favor of your proposal. But OTOH the behavior you observed is quite
> > clearly a bug, which we should fix independently of new updates to
> > the widget language. I think in this case we just need to be smarter
> > about calculating 'full_len' ...
> 
> I've thought about being smarter about calculating 'full_len', but
> unless we keep a cache of all other stuff drawn and then try to avoid
> drawing over that there is just no way to know the correct full_len,
> which is why I've rejected that approach.

That seems unnecessary complicated. It think we can calculate a
useful 'full_len' by something like
'w->length / display_props->cellwidth + 
	(w->length % display_props->cellwidth ? 1 : 0)'

> > Also when updating the widget language, we should think about bars, that
> > include text descriptions.
> 
> Well that would be a whole new type of bar IMHO and thus outside the
> scope of this PR.

Yes, outside the scope of the PR, but not outside the scope of #74.
I don't want to pump the protocol version with every minor change.
That would be hard to manage in client code. Thus we should do all
desired changes at once.

jwrdegoede:
On 18-04-19 11:02, haraldg wrote:
> AFAIR the command language is described in docs/LCDd.8.in and
> docs/lcdproc-dev ...

Ah yes, I would not have found it there myself. I will update the PR
to include an update to the docs.

> Yes, clients should check the protocol version reported by the server
> and use the proper set of commands (or give up).

If I understand things correctly, protocol version bumps are intended
mostly for changes which break compatibility. As I mentioned before I
deliberately made the changes in this PR in such a way that backward
compatibility is fully preserved. So I believe that in this case it
is best to not bump the version.

>  > I've thought about being smarter about calculating 'full_len', but
>  > unless we keep a cache of all other stuff drawn and then try to avoid
>  > drawing over that there is just no way to know the correct full_len,
>  > which is why I've rejected that approach.
> 
> That seems unnecessary complicated. It think we can calculate a
> useful 'full_len' by something like
> 'w->length / display_props->cellwidth +
> (w->length % display_props->cellwidth ? 1 : 0)'

That means that the length of the bar passed to the driver grows as
the bar gets filled more. This fixes the overwriting of stuff behind the
bar problem, so I think that this is a useful change in itself.

But this does not fix the problem of the driver being passed in incorrect
length, only now the length will often be too small rather then too big,
which for drivers which actually draw a "background" for the non filled
part of the bar, see:
https://fedorapeople.org/~jwrdegoede/cpu.png

Your suggestion would make the non-filled part of the bar slowly move to
the right as the filled part grows.

>  > Well that would be a whole new type of bar IMHO and thus outside the
>  > scope of this PR.
> 
> Yes, outside the scope of the PR, but not outside the scope of #74.
> I don't want to pump the protocol version with every minor change.
> That would be hard to manage in client code. Thus we should do all
> desired changes at once.

I understand that there are various big changes which you want to
make, but I have the feeling that neither of us has the time to
make those changes anytime soon. So IMHO it would be better to make
backward compatible incremental improvements where possible.

In case of the new 4 parameter set_widget syntax for bars, the client
can figure out if this is supported by trying it and if it is not
supported the client will get a "huh? Wrong number of arguments"
response.

Alternatively we could add a new "get_features" command which returns
a white-space separated list of features the server supports starting
with the word features, e.g. :

features featurename1 featurename2 feature3

This would be a new first patch in this patch-set and then the
commit adding the new 4 param set_widget variant for bars would
add a feature to the list called bar_widget_data_4_args and
clients which want to optionally use this can then check for it
using the new features command.

Thinking more about this, and since I'm arguing to allow making
small incremental improvements, I think such a new features command
would be quite useful to have and clients can use this as part
of their initial handshake, waiting for the reply, like they
do with the hello, without the client needing to check replies
once it is past the initial handshake.

This way we can keep the proto-ver at 0.3 and incrementally fix
things / add features, possibly also addressing some of the
things from #74.

jwrdegoede:
Thinking some more about the get_features idea, we could even make some of the features only show up depending on driver capabilities.

I know from the discussion in #74 that you are not in favor of adding support for graphics / pixel addressable LCDs and I agree that this is in general not a good fit.

But I've been thinking about eventually adding a new command to send a pixmap (which must be the exact size of the LCD resolution) to the LCD for LCDs which are capable of this.
This gives clients access to the graphics capabilities if they want to use them without lcdproc / LCDd itself getting involved other then passing along the pixmap.

Having the get_features command and showing some of the features only if the driver is capable of the feature would tie in nicely with this, since it would allow clients to discover if the LCD supports receiving pixmaps.

I've been toying with contacting Valve to see if they are interested in some sort of standard API for games to get access to (keyboard) LCD panels under Linux.
I think having graphics capabilities is going to be a must-have feature for the Valve/Steam use-case.

haraldg:
ad get_features: There is already the 'info' command. We currently are
lacking some documentation/standardization, but should be easy to
work on this if desired.

ad pixmaps: There is the icon widget. Currently it only supports
predefined icons. Maybe the possibility of uploading custom icons
could be considered, though I'm not quite sold on the idea yet.

I guess the main question is the usecase: If it's just to have some
API to control graphical displays, then probably something like
the fbtft kernel driver is better suited. If people really want to
have their graphical screen rotate with other lcdproc screens, then
I can see the point of adding this to LCDd. Of course if people want
this, I won't block it, but it feels rather ad hoc to me, so I'm not
quite convinced it's worthwhile.

> If I understand things correctly, protocol version bumps are intended
> mostly for changes which break compatibility. As I mentioned before I
> deliberately made the changes in this PR in such a way that backward
> compatibility is fully preserved. So I believe that in this case it
> is best to not bump the version.

We can certainly add new widgets or similar without the bump. But we
kind of can't use them in the (official) clients, I think: While surely
most users will use compatible versions of server and clients, the
interface is a network protocol and I believe running server and client
on different machines is supported. I have no idea how common this is,
but I don't feel comfortable breaking setups, where the client gets
upgraded but the server not (yet).

> Your suggestion would make the non-filled part of the bar slowly move
> to the right as the filled part grows.

Yes, indeed. I see your point. This kind of hbar, that is actually a
percentage instead of simply a width of pixels is not supported ATM.
It's reasonable to add this. I'm just not fond of breaking already
supported use cases for this.

> I understand that there are various big changes which you want to
> make, but I have the feeling that neither of us has the time to
> make those changes anytime soon.

Actually most of the proposed changes aren't that big and would have
been done long ago, if I was willing to do a protocol bump for minor
cleanups.

The actually big change would be liblcdproc, which makes handling
different versions much easier. You are right, that won't happen
anytime soon without somebody helping. I don't propose we wait on that.

> In case of the new 4 parameter set_widget syntax for bars, the client
> can figure out if this is supported by trying it and if it is not
> supported the client will get a "huh? Wrong number of arguments"
> response.

Not without significantly rewriting the clients and/or implementing
the proposed 'sync' command.

There are many ways forward and I'm not against anything as long as
* somebody does the work
* we don't break usecases supported for decades
* we don't turn our code base into more of a mess than it already is.

jwrdegoede:
On 18-04-19 15:35, haraldg wrote:
> ad get_features: There is already the 'info' command. We currently are
> lacking some documentation/standardization, but should be easy to
> work on this if desired.

This seems to currently just print the driver name, which is a free
form string. We could append "features: ..." at the end I guess
and the client can do a strstr for that.

I take it you would prefer that over adding a new get_features command?

> ad pixmaps: There is the icon widget. Currently it only supports
> predefined icons. Maybe the possibility of uploading custom icons
> could be considered, though I'm not quite sold on the idea yet.
> 
> I guess the main question is the usecase: If it's just to have some
> API to control graphical displays, then probably something like
> the fbtft kernel driver is better suited. If people really want to
> have their graphical screen rotate with other lcdproc screens, then
> I can see the point of adding this to LCDd. Of course if people want
> this, I won't block it, but it feels rather ad hoc to me, so I'm not
> quite convinced it's worthwhile.

The idea is to have a standard interface for e.g. games to get access
to LCD panels, where the client should always support using the standard
text APIs, but in case of games, may want to switch to graphical mode
if supported.

>  > If I understand things correctly, protocol version bumps are intended
>  > mostly for changes which break compatibility. As I mentioned before I
>  > deliberately made the changes in this PR in such a way that backward
>  > compatibility is fully preserved. So I believe that in this case it
>  > is best to not bump the version.
> 
> We can certainly add new widgets or similar without the bump. But we
> kind of can't use them in the (official) clients, I think: While surely
> most users will use compatible versions of server and clients, the
> interface is a network protocol and I believe running server and client
> on different machines is supported. I have no idea how common this is,
> but I don't feel comfortable breaking setups, where the client gets
> upgraded but the server not (yet).

Ok, understood. So before using the new 4 parameter set_widget calls on bars,
clients should first do a feature check.

>  > Your suggestion would make the non-filled part of the bar slowly move
>  > to the right as the filled part grows.
> 
> Yes, indeed. I see your point. This kind of hbar, that is actually a
> percentage instead of simply a width of pixels is not supported ATM.
> It's reasonable to add this. I'm just not fond of breaking already
> supported use cases for this.

I agree 100% that we should not break any existing use cases.

>  > In case of the new 4 parameter set_widget syntax for bars, the client
>  > can figure out if this is supported by trying it and if it is not
>  > supported the client will get a "huh? Wrong number of arguments"
>  > response.
> 
> Not without significantly rewriting the clients and/or implementing
> the proposed 'sync' command.

Right, which is why I suggested the feature flag approach, I assume
clients already wait; or are supposed to wait for the hello answer
to check the version. So they could do the same for the info command
and then after that continue as they currently do, so continue without
waiting for server answers / without syncing.

I've a feeling it might be easier to discuss some of this in real-time.
Are you on IRC? I'm hansg on freenode, you can find me in #dri-devel,
#fedora-workstation and other channels.

haraldg:
Hans de Goede writes:
> On 18-04-19 15:35, haraldg wrote:
> > ad get_features: There is already the 'info' command. We currently are
> > lacking some documentation/standardization, but should be easy to
> > work on this if desired.
> 
> This seems to currently just print the driver name, which is a free
> form string. We could append "features: ..." at the end I guess
> and the client can do a strstr for that.

It's up to the driver what it reports here. We sure improve this by
adding some standardization.

> I take it you would prefer that over adding a new get_features command?

Depends on what exactly get_features would do.
Right now we have:
protocol version: Tells what the server can do
info command: supposedly tells what the driver can do, but never really
              took off.

Ie the idea is there but nobody ever did much with it.

We can add yet an other feature reporting facility, though I'm not sure
what the point would be. I'm open to suggestions.

> > ad pixmaps: There is the icon widget. Currently it only supports
> > predefined icons. Maybe the possibility of uploading custom icons
> > could be considered, though I'm not quite sold on the idea yet.
> > 
> > I guess the main question is the usecase: If it's just to have some
> > API to control graphical displays, then probably something like
> > the fbtft kernel driver is better suited. If people really want to
> > have their graphical screen rotate with other lcdproc screens, then
> > I can see the point of adding this to LCDd. Of course if people want
> > this, I won't block it, but it feels rather ad hoc to me, so I'm not
> > quite convinced it's worthwhile.
> 
> The idea is to have a standard interface for e.g. games to get access
> to LCD panels, where the client should always support using the standard
> text APIs, but in case of games, may want to switch to graphical mode
> if supported.

My point is: Do you want to have screen rotation in the games case?
If not, then the linux kernel already has a standard interface, so
there is little point in inventing a new one. Also: I don't quite
believe lcdproc is in the position to set standards ... ;)

> >  > In case of the new 4 parameter set_widget syntax for bars, the client
> >  > can figure out if this is supported by trying it and if it is not
> >  > supported the client will get a "huh? Wrong number of arguments"
> >  > response.
> > 
> > Not without significantly rewriting the clients and/or implementing
> > the proposed 'sync' command.
> 
> Right, which is why I suggested the feature flag approach, I assume
> clients already wait; or are supposed to wait for the hello answer
> to check the version. So they could do the same for the info command
> and then after that continue as they currently do, so continue without
> waiting for server answers / without syncing.

Yes, that's more or less the idea.

Note that I'm not against bumping the protocol. Just saying we should
take a moment to figure out what else we want to change at the same
time. Also given the time we take between releases "incremental approach"
might not be all that incremental anyway.

> I've a feeling it might be easier to discuss some of this in real-time.
> Are you on IRC? I'm hansg on freenode, you can find me in #dri-devel,
> #fedora-workstation and other channels.

Sorry, I'm not on IRC. And actually like to check up on the code or
documentation before replying. I only took on lcdproc, because it
is mostly unmaintained. I'm not that familiar with the code base.
In a way I'm just acting release mangager until somebody more dedicated
shows up/returns and don't actually have powers like adding people etc.

jwrdegoede:
About supporting sending pixmaps to graphics capable LCDs that is a bit off-topic for this PR really, so lets park that discussion for now.

About the issues with (h) bar widgets currently overwriting stuff to the right of them (on the mx5000 at least). Based on our discussion on this my plan is to:

1. Modify the info command so that it reports a list of supported optional features to the client.
2. Add (optional) support for a new hprogressbar widget with as parameters x, y, width and promillage, this widget draws a progress-bar with background for the non filled part like the current mx5000 hbar code does, versus the existing hbar widget which simply draws a horizontal bar of length pixels long. Support for the hprogressbar will be advertised as a feature by the new info command
3. Drop the current hbar implementation from the mx5000 driver, as it is actually a hprogressbar not a hbar
4. Teach the lcdproc client to get a list of supported features
5. Add hpercentbar_init and hpercentbar_update helper functions to the lcdproc client, which widget_add / widget_set a hprogressbar if the feature is advertised and otherwise they fallback to using a good old hbar
6. Modify some of the lcdproc screens to use the new hpbar_init and hpbar_update helpers

Before I start reworking my patches according to this plan, does this sound like it is going in the right direction?

jwrdegoede:
So I've been looking into step 1 a bit, the info command is not even documented in the LCDd manpage I wonder if we can count on older servers having some form of it.

Also at least for the lcdproc client, needing to send 2 commands seems to be tricky to build into the state-machine.

So I think a better way to pass features is by adding " feat feature-name" 1 or multiple times at the end of the connect string, I've checked all existing default clients and a couple of ones (with non broken links) from http://lcdproc.org/clients.php3 and they are fine with adding extra text at the end of the connect string.

haraldg:
You are proposing to put quite some work into this code. While this is
always welcome, I'm a bit surprised to see that much effort going into
what appears to me as an obscure and rare piece of HW. Especially since
I guess that this is part of paid work rather than a hobby. I wonder
what the motivation and overall goal is. How many users do you expect
lcdproc will have on this HW? (Not that I need to know. Just curious.)

If you are willing to do this kind of serious development on lcdproc,
I recommend you take the time to read most parts of our developers guide.
I hope this will - among other thinges - help you understand how
everything in lcdproc is meant to fit together.

The important thing is, that we try to have a clear abstraction of
layers: clients <-> server <-> drivers

We already talked some about the protocol between clients and server.
Between server and drivers there is similarly a defined and versioned
API. Also there is the expectation that any feature, not natively
supported by the driver in use, is emulated by the server in some basic
way. (The easiest way to fix the mx5000 hbars actually would be to
just drop the custom implementation and use the server side emulation.
The result would be less aestetic, but more in line with the usual
lcdproc experience. Maybe this is worth considering for you.)

Actually there are two conflicting design goals here, that cause some
of the mess in our code base: On one hand we try to provide a fairly
abstract widget language, that makes writing clients easy: "Just tell
the server which widgets you want and let it handle all the details".
On the other hand given the sever size constraints of the typically
used displays, the clients need/want to optimize their output a lot make
good use of the valuable space.

To answer your points specifically:

ad 1) Yes, info never really took off, but it has been around for at
least 17 years - probably longer. Adding additional information to the
connect string would be possibly I guess - though I fail to see the
difference to bumping the protocol version.

I think the point is, that the connect string is about the server and
the info command is about the (display) driver. (Thinks like are
hbars character based or pixel based, which icons are available, etc.
could be reported here.)

ad 2) I don't think we want optional widgets. If we add a new widget,
we should add a default implementation to the server like I have explained
above.

I don't have a strong opinion if the "new" widget should get a new
name or just be a hbar with 4 parameters. I can thing of good reasons
either way. However I'd suggest to use something shorter then "hprogressbar".
The 'h' seems rather superfluous and also this substring will be part of
many lines of code. Maybe use "pbar" or even just "bar".

A new widget is a server feature, not a driver feature, so I think it
should be advertises in the connect string, not the info command.

Some consideration will be necessary how to make this work with the
driver API. I guess we will need a new callback and drivers would register
either their hbar implementaion either for the old hbar or the new
bar or even for both. I haven't thought about this in detail.

add 3) Actually this can be done at any time without other changes.
Legacy hbars will remain being available from the fallback implementation.

add 4) Sure. Preferably without inventing new ABI (or at least without
adding complexity to the ABI).

add 5/6) That's basically what I'm heading for with liblcdproc too, so I
guess I like that.

jwrdegoede:
On 19-04-19 11:25, haraldg wrote:
> You are proposing to put quite some work into this code. While this is
> always welcome, I'm a bit surprised to see that much effort going into
> what appears to me as an obscure and rare piece of HW. Especially since
> I guess that this is part of paid work rather than a hobby.

This is a hobby / weekend project for me, no (or very little) paid time
is going into this.

> I wonder
> what the motivation and overall goal is. How many users do you expect
> lcdproc will have on this HW? (Not that I need to know. Just curious.)

As I mentioned in #116 I've been working on a kernel patch series to
improve support for Logitech wireless kbds / mice, see:
https://www.spinics.net/lists/linux-input/msg60944.html

I bought a 2nd hand MX5000 to be able to test this series with a
Logitech BT receiver in HID proxy mode. Then as a hobby project I
tried to get the LCD to work. It sorta worked using existing libmx5000
code but it looked quite bad which is how we ended here.

Basically I like tinkering with hardware and getting hw to work
100% with Linux. I hope that other MX5000 owners will benefit from
this too (eventually), how much Linux MX5000 users there are I have
no idea.

TL;DR: this is all just a hobby project, born from me liking to tinker.

> The important thing is, that we try to have a clear abstraction of
> layers: clients <-> server <-> drivers

I understand.

> We already talked some about the protocol between clients and server.
> Between server and drivers there is similarly a defined and versioned
> API.

This versioned API seems to be something which was intended to at
one point in time allow drivers outside of the lcdproc source-tree,
but I do not think this has ever materialized.

Specifically it seems that all drivers depend on server/drivers/lcd.h
and make install never installs that file under /usr/include, so there
can really not be any out of tree drivers.

As such I believe it is fine to add extra symbols under:

         /* extended output functions (optional; core provides alternatives) */

Without worrying about this changing the struct layout, since the drivers
are build at the same time as LCDd, so they will use the same version
of the struct. Much like how the kernel and kernel modules have a flexible
API between the kernel and the modules as they too are always build from
the same source tree.

But if you are worried about ABI compatibility between LCDd and the driver
.so files then we can just add new "extended output functions" members at
the end of the struct, the struct is always allocated by LCDd itself
(in driver_load()), so we can safely grow it without breaking ABI and
driver_bind_module() will happily keep loading old modules without
the (optional) symbol for, for example, a new pbar output function,
since it already supports optional symbols.

> Also there is the expectation that any feature, not natively
> supported by the driver in use, is emulated by the server in some basic
> way.

Ok, that can easily be done with the new pbar, it can just fallback
to using a hbar instead.

> (The easiest way to fix the mx5000 hbars actually would be to
> just drop the custom implementation and use the server side emulation.
> The result would be less aestetic, but more in line with the usual
> lcdproc experience. Maybe this is worth considering for you.)

Well I already have the mx5000 working in what IMHO is a much nicer
way then falling back to the fallback --- hbar implementation, so
I would prefer to get what I currently have working in a way which
is acceptable to you. And I believe that this will also allow using
nicer bars on various other LCDs.

Besides getting some notion of feature flags into the LCDd protocol
is something which will be useful for other incremental improvements
in the future so this seems like it is something worth working on.

> ad 1) Yes, info never really took off, but it has been around for at
> least 17 years - probably longer. Adding additional information to the
> connect string would be possibly I guess - though I fail to see the
> difference to bumping the protocol version.
> 
> I think the point is, that the connect string is about the server and
> the info command is about the (display) driver. (Thinks like are
> hbars character based or pixel based, which icons are available, etc.
> could be reported here.)

Ok, so if I add support to the server for a new pbar widget and we want
clients to know about this, then this should be added to the connect string.

> ad 2) I don't think we want optional widgets. If we add a new widget,
> we should add a default implementation to the server like I have explained
> above.

Ok, we still need a way for a client to find out if the server is new
enough to support the widget though.

> I don't have a strong opinion if the "new" widget should get a new
> name or just be a hbar with 4 parameters. I can thing of good reasons
> either way. However I'd suggest to use something shorter then "hprogressbar".
> The 'h' seems rather superfluous and also this substring will be part of
> many lines of code. Maybe use "pbar" or even just "bar".

Well the hbar is really a horizontal bar starting at x y of pixels width
large (from the server API pov, the driver API is not a good match to the
hbar server API). So I think that if want a progress / percentage bar,
with clearly different behavior then a new widget is best.

> Some consideration will be necessary how to make this work with the
> driver API. I guess we will need a new callback and drivers would register
> either their hbar implementaion either for the old hbar or the new
> bar or even for both. I haven't thought about this in detail.

See above, we can simply add a new pbar() function to the Driver struct,
as said if you are worried about LCDd and driver.so ABI compat we can
safely add this at the end and old drivers will keep working without
needing to be recompiled.

> add 3) Actually this can be done at any time without other changes.
> Legacy hbars will remain being available from the fallback implementation.
> 
> add 4) Sure. Preferably without inventing new ABI (or at least without
> adding complexity to the ABI).
> 
> add 5/6) That's basically what I'm heading for with liblcdproc too, so I
> guess I like that.

Ok.

haraldg:
We are pretty liberal about merging drivers, so there isn't much
out of tree stuff. Though there are unmerged drivers because of
legal or technical reasons. I have no idea how they are being built.
Some people love manual work ... :)

But anyway it's not a big concern. We don't care much about
server-driver ABI compatibility. And we have broken the API in
minor ways in the past without bumping the version. (Mostly to get
rid of some of the more overengineered stuff.)

Still I try to keep the API clean (or even clean it up a little).
I don't see any problem with what you propose. I'm just pointing
out that I haven't thought about all the details as a disclaimer
to my general agreement.

> Ok, that can easily be done with the new pbar, it can just fallback
> to using a hbar instead.

It's one possiblity. Though somehow indicating the unfilled part
of the pbar would be nice.

> Besides getting some notion of feature flags into the LCDd protocol
> is something which will be useful for other incremental improvements
> in the future so this seems like it is something worth working on.

I won't block this, but I still don't see why we wouldn't just pump
the version. Maybe I'm overly cautious, but maintaining this codebase,
which I believe is older then myself, has taught me. (You increase the
length of a line somewhere and suddenly somewhere else it gets truncated
because some buffer is too short. That kind of stuff.)

jwrdegoede:
On 19-04-19 15:27, haraldg wrote:
>  > Besides getting some notion of feature flags into the LCDd protocol
>  > is something which will be useful for other incremental improvements
>  > in the future so this seems like it is something worth working on.
> 
> I won't block this, but I still don't see why we wouldn't just pump
> the version.

Bumping the version to 0.4 for the addition of the pbar widget sounds
like a good solution to me, but I thought that you only wanted to do
that when it is paired with some other big changes, If you're fine
with doing the bump for the pbar addition then I agree that that
is the best solution.

haraldg:
Yes, if we bump the version I will try to get the other planned changes
in before the next release. I don't think they are particularly big or
difficult.

jwrdegoede:
On 20-04-19 12:27, haraldg wrote:
> Yes, if we bump the version I will try to get the other planned changes
> in before the next release. I don't think they are particularly big or
> difficult.

Ok, lets go with the version bump then. I do not have much time to
work on this this weekend. I will get back to you with an update PR
when I'm done with the changes.

haraldg:
Sure. Feel free to submit incremental (but self contained, including
documentation) PRs. The time I have for lcdproc varies a lot. A big PR
might take quite some time for review on my end...

jwrdegoede:
> Sure. Feel free to submit incremental (but self contained, including documentation) PRs.

Ok I've just pushed a new set of commits to this branch / this PR implementing steps 1-3 of the discussed plan. I will submit steps 4-6 as a separate PR based on top of this one.

jwrdegoede:
I've made nice progress today with steps 4-6, see the last 3 commits here: https://github.com/jwrdegoede/lcdproc/commits/master taking a peek at this is useful to get an idea of how the new pbar widget can be used.

I've tested the converted lcdproc CPU screen on both 4 and 2 line mode and it works well (I also forced fallback on both client and server side for testing). I especially like how the [ ] go away on the 2 line mode when the driver has a native progress/percentage-bar, giving us significantly more space for the bar. This will be a nice improvement as more drivers (hopefully) get support for this in the future.

I still want to convert the mem screen at a minimum. I likely will not get around to doing that until next weekend.

server/commands/widget_commands.c
```diff
@@ -296,6 +296,35 @@ widget_set_func(Client *c, int argc, char **argv)
...
+		w->x = atoi(argv[i]);
+		w->y = atoi(argv[i + 1]);
+		w->width = atoi(argv[i + 2]);
+		w->promille = atoi(argv[i + 3]);
+		if (argc == i + 6) {
+			w->begin_label = strdup(argv[i + 4]);
+			w->end_label = strdup(argv[i + 5]);
+			if (!w->begin_label || !w->end_label) {
+				report(RPT_WARNING, "widget_set_func: Allocation error");
+				return -1;
```
haraldg: I wouldn't bother checking allocation here:
* if this ever fails, we have bigger trouble, then that
* nothing really bad happens when a label is missing
* it leads to questions what to answer the client

Genereally I believe allocation of small data structures should be treated as fail safe in userspace applications. (Excecpt for very rare corner cases.)

server/driver.c
```diff
@@ -447,6 +448,56 @@ driver_alt_hbar(Driver *drv, int x, int y, int len, int promille, int options)
...
+	end_length = end_label ? strlen(end_label) : 0;
+
+	/* We want at least 2 chars for the bar itself */
+	if ((begin_length + end_length + 2) > width)
+		begin_length = end_length = 0;
+
+	len = width - begin_length - end_length;
+
+	for (i = 0; i < begin_length; i++)
+		drv->chr(drv, x++, y, begin_label[i]);
```
haraldg: Any reason to use drv->chr over drv->string?

server/drivers.c
```diff
@@ -309,6 +309,40 @@ drivers_hbar(int x, int y, int len, int promille, int pattern)
...
+ * space.
+ */
+void
+drivers_pbar(int x, int y, int width, int promille, char *begin_label, char *end_label)
+{
+	Driver *drv;
+
+	debug(RPT_DEBUG, "%s(x=%d, y=%d, width=%d, promille=%d)", __FUNCTION__, x, y, width, promille);
+
+	ForAllDrivers(drv) {
```
haraldg: Not your fault, but this is just horrible ...

server/widget.c
```diff
@@ -85,7 +86,7 @@ widget_create(char *id, WidgetType type, Screen *screen)
 	debug(RPT_DEBUG, "%s(id=\"%s\", type=%d, screen=[%s])", __FUNCTION__, id, type, screen->id);
 
 	/* Create it */
-	w = malloc(sizeof(Widget));
+	w = calloc(1, sizeof(Widget));
```
haraldg, line 89: Is there a reason we need initialzed memory here?

haraldg (changes requested):
A number of nits inline. More importantly:

You need to add proper documentation to the developers guide.

I think begin/end labels have unclear semantics. On one hand you use them in the default implementation to mark the width of the widget. On the other hand you treat them as general purpose labels. In the end I'm not sure, why they are there at all. Let's break this issue down into separate topics:

The labels in the pbar driver API: I think the drivers provide primitive operations to the server, not full implementations of actual widgets. I don't think rendering labels as part of the pbar driver operation makes sense, unless the labels are intended to be rendered within the bar itself (using inverse color), which couldn't be done by the server. As it is now, this only leads to code duplication in the drivers.

The lables in the widget language might make sense, if there are lot's of screens where such labels are used and would be added directly otherwise. However in that case it doesn't make sense to have them default to brackets. The default should be empty strings. 

In any case the brackets in the default implementation should be treated differently from labels.

jwrdegoede:
On 25-04-19 15:16, haraldg wrote:
> You need to add proper documentation to the developers guide.

Ok, will fix.

> I think begin/end labels have unclear semantics. On one hand you use them in the default implementation to mark the width of the widget. On the other hand you treat them as general purpose labels. In the end I'm not sure, why they are there at all. Let's break this issue down into separate topics:
> 
> The labels in the pbar driver API: I think the drivers provide primitive operations to the server, not full implementations of actual widgets. I don't think rendering labels as part of the pbar driver operation makes sense, unless the labels are intended to be rendered within the bar itself (using inverse color), which couldn't be done by the server. As it is now, this only leads to code duplication in the drivers.
Ok, leaving the rendering of the labels to the server instead of doing it in the drivers is fine with me, I will adjust this for the next version.
> The labels in the widget language might make sense, if there are lot's of screens where such labels are used and would be added directly otherwise. However in that case it doesn't make sense to have them default to brackets. The default should be empty strings.
There are 2 different paths here:

1. The driver supports the pbar method, so there are clear begin and end markings of the pbar drawn by the driver. I agree that in this case the default begin- and end-labels should be an empty string and this is what the mx5000 driver does.

2. The driver does *not* support the pbar method, in this case if there are no labels we need something to mark the beginning and end of the area the hbar uses. Currently the 2 line lcdproc cpu and mem screens:

```
+--------------------+  +--------------------+
|CPU [----    ]48.1%@|  |M 758.3M [- ] 35.3%@|
|U--  S-   N    I--- |  |S 1.884G [  ]  0.1% |
+--------------------+  +--------------------+

```

Use [] around the hbar in this case, so the idea is to have the default/fallback implementation draw the [] as begin and end marking, where as in the case the driver supports the pbar method these are left out, gaining us 2 extra chars in width for the actual pbar.

IOW the idea is that that if no begin/end labels are provided and we are using a hbar to "emulate" a pbar then we insert begin/end markers ourselves and I've chosen [] for this so that there is no functional chance when the 2 line cpu and mem screens switch from drawing '[' hbar ']' to drawing a label less pbar.

I hope this explains why I believe that the behavior which I've implemented is correct / desirable.

server/driver.c
jwrdegoede: I'm not sure that *all* drivers implement string, the check above is for driver->chr, not drv->string.

server/widget.c
jwrdegoede: Otherwise the new begin_label, end_label and promille fields explicitly set to NULL/0. I'm used to just using calloc so everything in the struct has a known value (NULL/0) and only when a different value is used explicitly initialize things. If you prefer I can add explicit initialization of the 3 new members.

server/commands/widget_commands.c
jwrdegoede: Ok, I will drop the allocation success checks.

server/widget.c
haraldg: It's ok. I was just confused, because it looked like an unrelated change.

haraldg:
All you write makes sense to me except for the conclusion: If the labels are just for the fallback implementation, why have them as parameter in the widget language at all?

The client (probably) doesn't know if the display uses native pbar or fallback. So why would it specify the markers at all? I think the fallback implementation can just use [] at all times ...

jwrdegoede:
On 28-04-19 22:21, haraldg wrote:
> All you write makes sense to me except for the conclusion: If the labels are just for the fallback implementation, why have them as parameter in the widget language at all?
> 
> The client (probably) doesn't know if the display uses native pbar or fallback. So why would it specify the markers at all? I think the fallback implementation can just use [] at all times ...

The [] should only be used if there are no other start/end markers like "0%" and "100%"; or "E" and "F", this is the main reason to make the labels part of the widget language, this way the fallback implementation can decide to add the [] markers based on the lack of other begin and end-labels.

Another reason to have the labels in the widget language is so that if a fancy LCD does support rendering them inside the bar somehow that we can then add support for that without modifying the widget language again.

Also since most clients seem to often use labels, having the labels as part of the widget language makes writing clients easier, since they now do not need to calculate the label width and placement.

haraldg:
If there is an actual use case and this makes writing clients easier, then fair enough. However I still don't see how this will work well in both (native pbar and fallback pbar) cases using the semantics you propose. I guess I will need to wait for your updated implementation to see where you are taking this.

jwrdegoede:
Ok, I've just pushed a new version addressing all review remarks I hope.

For an idea of how the new pbar widget can be used, see my master branch which converts the lcdproc cpu screen (other screens to follow soon): https://github.com/jwrdegoede/lcdproc/commits/master

server/driver.c
jwrdegoede: So I've checked and all output drivers do provide a string method. So I've switched to using drv->string here in the version which I've just pushed.

jwrdegoede:
Thank you for the detailed review and for merging this.

Issue #109, opened 2018-10-03
switch from libugpio to libgpiod


haraldg:
The old sysfs gpio interface of the linux kernel is deprecated in favor of a new ioctl based API. Thus libugpio is discontinued. There is a new library libgpiod that makes interfacing the new API easier. Thus we should switch from one library to the other.

Since the addressing scheme of GPIOs is entirely different now, also driver configuration well need changes, but since the libugpio based connection type never saw much use (likely my bad for pushing libugpio shortly before it was deprecated) this shouldn't be much of a problem.

I kind of feel responsible for this situation and will look into it during the cold season. But if somebody wants to work on it sooner, just leave a note here an feel free to take it over. Should be fairly easy and beginner friendly topic to work on!

PR #119, opened 2019-04-18
hd44780-gpio: switch from libugpio to libgpiod

haraldg:
Hi,

I finally got around to do the port to the new API. However I don't have 
any HW to test this on. Can somebody with a hd44780 connected to
GPIOs test and report back?

On systems with only one gpio controller only a single line (the
gpiochip parameter) needs to be added to LCDd.conf. On systems with
multiple gpio controllers the gpio numbers need to be edited carefully.
However such systems weren't properly supported in the past anyway.

This closes issue #109.

Signed-off-by: Harald Geyer <harald@ccbib.org>


haraldg:
@antonio-petricca @JamesGKent @ElFishi @terba you are the people reporting issues with the gpio connection type in the past. Can you test this PR for me? I don't want to merge this without any testing at all.

terba:
I don't have hardware to test it quickly and due to lack of time can not build one now. Sorry.

antonio-petricca:
Hi, unfortunately I cannot test it because I do not have anymore the hardware to perform the check.

JamesGKent:
I'm currently without internet at home where my devices are and have been for a couple of weeks, I can test once its restored with a 4x40 screen.

ethandicks:
I just got a Pi4 this week.  I can hook an HD44780 to it and give this a try.

mskalski:
I tried this in April and I could't compile it right. Problem is that many Linux distributions (still supported, i.e Ubuntu 16.04) have old kernel, so it would be not possible to use `libgpiod`, on the other hand `libugpio` is supported.

I would suggest to leave possibility to use `libugpio`, and add `libgpiod` as another option, possibly mutually exclusive, but I would prefer as independent options.

haraldg:
@mskalski thanks for your feedback. 

I have been thinking, that libugpio never was available on systems like Ubuntu 16.04, so it wouldn't be a problem. If people actually are selfcompiling lcdproc and libs on systems with old kernels or vendor kernels, then we shouldn't actively break this at least.

ethandicks:
It took me a lot longer to put the hardware together than I planned, but I finally have a working 20x4 LCD on a 40-pin Raspberry Pi GPIO cable.  I was not able to get the code to compile as given but was able to get it fully working and tested on a Raspberry Pi 4B with the following trivial changes:

`
$ git diff
diff --git a/server/drivers/hd44780-drivers.h b/server/drivers/hd44780-drivers.h
index d03a31a..06434f8 100644
--- a/server/drivers/hd44780-drivers.h
+++ b/server/drivers/hd44780-drivers.h
@@ -45,9 +45,10 @@
 #ifdef WITH_RASPBERRYPI
 #include "hd44780-rpi.h"
 #endif
-#ifdef HAVE_UGPIO
+#ifdef HAVE_GPIOD
 #include "hd44780-gpio.h"
 #endif
+
 /* add new connection type header files to the correct section above or here */
 
 
@@ -104,7 +105,7 @@ static const ConnectionMapping connectionMapping[] = {
 #ifdef WITH_RASPBERRYPI
        { "raspberrypi",   HD44780_CT_RASPBERRYPI,   IF_TYPE_PARPORT,  hd_init_rpi      },
 #endif
-#ifdef HAVE_UGPIO
+#ifdef HAVE_GPIOD
        { "gpio",          HD44780_CT_GPIO,          IF_TYPE_PARPORT, hd_init_gpio      },
 #endif
        /* add new connection types in the correct section above or here */
`

haraldg:
Good catch. I guess, I didn't notice this during compile testing, because I had both libraries installed on my system back then.

I'll submit a new PR addressing your feedback and that of @mskalski. I guess, we will end up with an ugpio and gpiod connection type. :-(

ethandicks:
I suppose adding gpiod support rather than replacing ugpio with gpiod will avoid breaking old LPT-attached devices but still get us in the right place for supporting all the varieties of Raspberry Pi.

haraldg:
Almost done reworking this. @ethandicks one last thing:
Can you share the section with the pin setup from you configuration file? I'd like to add an example for rpi/gpiod to the configuration.

(The existing example is actually ill suited for gpiod, because it uses gpio lines across different chips. Also the RPi seems more popular then the BB.)

ethandicks:
> Almost done reworking this. @ethandicks one last thing:
> Can you share the section with the pin setup from you configuration file? I'd like to add an example for rpi/gpiod to the configuration.

Can do.  I'll have to pull out the Pi400 and fire it up but that shouldn't take long.

ethandicks:
Here's the relevant part of my LCDd.conf from my testing:

ConnectionType=gpio

gpiochip=0
pin_D4=25
pin_D5=24
pin_D6=23
pin_D7=18
pin_EN=8
pin_RS=7
#pin_BL=xx

Obviously, 'gpiod' would be correct now.

haraldg:
Thanks, this has been helpful. The final change is in https://github.com/lcdproc/lcdproc/commit/3f9bc27f5802e43cc6732dc7409dd6f2e748ed4e

manio:
@haraldg 
I am currently using `ConnectionType=raspberrypi` wiring for the hd44780.
From what I understanding, to use this `gpio` connection I need:
- kernel which supports gpiod
- userspace `libgpiod` library installed
- compiled LCDproc driver with `gpiod` support

and config:
- change from raspberrypi to `ConnectionType=gpio`
- add `gpiochip=0` - is this mandatory?

The pin definitions stays the same?

PR #120, opened 2019-05-03
Clients lcdproc use pbar

jwrdegoede:
Convert the clients/lcdproc client's cpu and memscreen to use the new pbar widget.

### DIFF
--- a/clients/lcdproc/cpu.c
+++ b/clients/lcdproc/cpu.c
@@ -58,7 +58,7 @@ cpu_screen(int rep, int display, int *flags_ptr)
 	static int us_wid = 0;
 	static int ni_wid = 0;
 
-	int i, j, n;
+	int i, j;
 	double value;
 	load_type load;
 	char tmp[25];		/* should be large enough */
@@ -69,7 +69,6 @@ cpu_screen(int rep, int display, int *flags_ptr)
 		sock_send_string(sock, "screen_add C\n");
 		sock_printf(sock, "screen_set C -name {CPU Use: %s}\n", get_hostname());
 		if (lcd_hgt >= 4) {
-			gauge_wid = lcd_wid - 6;	/* room between 0%...100% */
 			us_wid = ((lcd_wid + 1) / 2) - 7; /* Usr/Sys label width -7 for " xx.x% " */
 			ni_wid = lcd_wid / 2 - 6;       /* Nice/Idle label width -6 for " xx.x%" */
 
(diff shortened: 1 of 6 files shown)


haraldg:
Looks mostly good.

I think there is a corner case in your pbar helpers: If the set helper
is first called with non-empty labels, but later called again for the
same widget with empty labels, the label won't be cleared. Ie what ends
up visible on the screen depends on the order in which widgets are
rendered. I think this can't happen with the current uses of the
helper, so this is no show stopper. But I felt I should point this
out. If nothing else, maybe at least add a warning in a comment.

Your display flag handling change actually changes behaviour. I'm
not against this - the code goes back to the dawn of version control
and it isn't really clear to me, what is intended. AFAICT it seems
to give the screen functions a way to track time while their screens
are not shown, but I can't say I have fully grasped it. Anyway, as your
proposed change stands, we might as well get rid of the display flag
and the show_invisible screen setting entirely. (I haven't looked what
this entails, but the new behaviour probably needs updates to the
documentation in various places.)

jwrdegoede:
On 04-05-19 14:46, haraldg wrote:
> Looks mostly good.

As always thank you for the review, much appreciated.

> I think there is a corner case in your pbar helpers: If the set helper
> is first called with non-empty labels, but later called again for the
> same widget with empty labels, the label won't be cleared. Ie what ends
> up visible on the screen depends on the order in which widgets are
> rendered.

Good point, I will rework the helper to set the string widgets to an empty string if the label-length is 0.

> Your display flag handling change actually changes behaviour.

I do not believe it does, the calls which are skipped purely read info from under /proc and store it in on stack variables.

So what this changes is instead of reading info from /proc, storing this on the stack and rendering some strings on stack; only to never output those strings to the socket, it simply bails.
This should not result in any functional change.

haraldg:
Hans de Goede writes:
> > Your display flag handling change actually changes behaviour.
> 
> I do not believe it does, the calls which are skipped purely read info from under /proc and store it in on stack variables.

It does. Please review mem.c again. Especially:

| /* flip the title back and forth... (every 4 updates) */
| if (which_title & 4)
|	sock_printf(sock, "widget_set M title {%s}\n", get_hostname());
| else
|	sock_printf(sock, "widget_set M title { MEM %.*s SWAP}\n", title_sep_wid, title_sep);
| which_title = (which_title + 1) & 7;

AFAIKS this code is the only reason, the display flag is even there.
There are a few other screens with similar code. Those have the
show_invisible property enabled by default. (see main.c)

jwrdegoede:
On 04-05-19 18:53, haraldg wrote:
> Hans de Goede writes:
>  > > Your display flag handling change actually changes behaviour.
>  >
>  > I do not believe it does, the calls which are skipped purely read info from under /proc and store it in on stack variables.
> 
> It does. Please review mem.c again. Especially:
> 
> | /* flip the title back and forth... (every 4 updates) */
> | if (which_title & 4)
> | sock_printf(sock, "widget_set M title {%s}\n", get_hostname());
> | else
> | sock_printf(sock, "widget_set M title { MEM %.*s SWAP}\n", title_sep_wid, title_sep);
> | which_title = (which_title + 1) & 7;
> 
> AFAIKS this code is the only reason, the display flag is even there.
> There are a few other screens with similar code. Those have the
> show_invisible property enabled by default. (see main.c)

Hmm, I see, it seems not useful though to keep flipping the title back and forth
when the screen is not visible.

Is there perhaps some menu screen somewhere which shows a list of all
screens? That would make this behavior somewhat useful...

Anyways I will rework the patch simplifying the flag handling to move
the code you quoted to above the:

'''
         if (!display)
                 return 0;
'''

So that there is no functional change was my intention.

I agree that we should probably remove the show_invisible and display flags
all together, but that falls outside the scope of this patch-set / pull-req.

Regards,

Hans

jwrdegoede:
Ok, the version which I've just pushed should address all your review remarks.

PR #121, opened 2019-05-03
G15 driver fixes

jwrdegoede:
After testing with a G15 keyboard I noticed various corruptions of the top and bottom rows of characters when using lcdproc with the G15 LCD screen.

This pull-req contains a series of refactoring + fixes patches which fixes this.

### DIFF
--- a/server/drivers/g15.c
+++ b/server/drivers/g15.c
@@ -94,6 +94,16 @@ MODULE_EXPORT int g15_init (Driver *drvthis)
 		return -1;
 	}
 
+	p->font = g15r_requestG15DefaultFont(G15_TEXT_LARGE);
+	if (p->font == NULL) {
+		report(RPT_ERR, "%s: unable to load default large font", drvthis->name);
+		g15_close_screen(p->g15screen_fd);
+		free(p->canvas);
+		free(p->backingstore);
+		free(p);
+		return -1;
+	}
+
 	g15r_initCanvas(p->canvas);
 	g15r_initCanvas(p->backingstore);
 	p->canvas->buffer[0] = G15_LCD_WRITE_CMD;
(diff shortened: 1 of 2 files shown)


server/drivers/g15.c
```diff
@@ -94,6 +94,16 @@ MODULE_EXPORT int g15_init (Driver *drvthis)
 		return -1;
 	}
 
+	p->font = g15r_requestG15DefaultFont(G15_TEXT_LARGE);
+	if (p->backingstore == NULL) {
```
haraldg: copy&paste error here

server/drivers/g15.c
```diff
@@ -202,61 +213,52 @@ int g15_convert_coords(int x, int y, int *px, int *py)
...
 
-	y--;
-	x--;
-	if ((x > p->width) || (y > p->height))
+	if (!g15_convert_coords(x, y, &px, &py))
 		return;
 
-	g15r_renderCharacterLarge(p->canvas, x, y, c, 0, 0);
+	/* Clear background */
+	g15r_pixelReverseFill(p->canvas, px, py, px + 7, py + 7, G15_PIXEL_FILL, G15_COLOR_WHITE);
```
haraldg: any reason you use literal 7s here instead of something involving G15_CELL_WIDTH/HEIGHT?

haraldg (changes requested):
Thanks for the cleanup and fixes!

server/drivers/g15.c
jwrdegoede: Good catch fixed for the next version.

server/drivers/g15.c
jwrdegoede: No special reason, I will switch to using G15_CELL_FOO - 1 for the next version.

jwrdegoede:
Ok, the version which I've just pushed should address all your review remarks.

PR #123, opened 2019-05-12
Client language: Update key handling

haraldg:
Hi all,

this is yet an other change where I'd like to have some input from the
community before pushing them to master. Of course code reviews
are always welcome, but even more important is proof reading the docs
(english is not my native language) and comments wether these
changes are a good idea at all.

The new commands allow to reserve keys per screen instead of per client.
This hopefully has clearer semantics then client_add_key in shared mode.
Also for clients with multiple screens this allows to have interaction
with just some of the screens while others still benefit from internal
key handling in the server.

TIA,
Harald

### DIFF
--- a/docs/lcdproc-dev/language.docbook
+++ b/docs/lcdproc-dev/language.docbook
@@ -403,6 +403,49 @@
 
 	<varlistentry>
 	  <term>
+	    <command>key_add
+	      <option><replaceable>screen_id</replaceable></option>
+	      <option><replaceable>key</replaceable></option>...
+	    </command>
+	    since protocol version 0.4
+	  </term>
+	  <listitem>
+	    <para>
+	      Reserves key(s) for interaction of the user with the screen.
+	      Reservations with this command always take precedence over
+	      reservations with <command>client_add_key</command>, but are
+	      only in effect while the screen is visible on the display.
+	      The server message on key events is appended with
+	      <replaceable>screen_id</replaceable> to distinguish it from
+	      client level key events.
+	    </para>
+
+	    <para>
+	      Only one success message is generated for all keys together.
+	    </para>
+	  </listitem>
+	</varlistentry>
+
+	<varlistentry>
+	  <term>
+	    <command>key_del
+	      <option><replaceable>screen_id</replaceable></option>
+	      <option><replaceable>key</replaceable></option>...
+	    </command>
+	    since protocol version 0.4
+	  </term>
+	  <listitem>
+	    <para>
+	      Lift reservation of key(s) by <command>key_add</command>.
+	      Arguments are processed sequentially and for each one either
+	      a success message is generated or an error message, if the
+	      key was not found in the list of reserved keys.
+	    </para>
+	  </listitem>
+	</varlistentry>
+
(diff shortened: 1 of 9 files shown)


haraldg:
Updated this PR to resolve conflicts. @ethandicks can you have a look, whether you consider these two changes good ideas.

And maybe also comment on how to improve my very austrian english to something, that's actually comfortable to read for native speakers?

ethandicks:
I will review this weekend.

ethandicks:
This change seems fine to me on sight (I did not compile or run it) and the English is fine.  About the only two very minor things in language.docbook that I spotted were:

1) The clauses on this are a little tangled because of how things get grouped together with the two conjunctions and the comma.  

      "Arguments are processed sequentially and for each one either
	      a success message is generated or an error message, if the
	      key was not found in the list of reserved keys."

What might group the two halves together better, idiomatically, is:

      "Arguments are processed sequentially, and for each one either
	      a success message is generated, or an error message if the
	      key was not found in the list of reserved keys."

There are probably several ways to rewrite this entirely, but by adding/moving the commas, I think it flows better.

2) Similarly, the pacing is just a little odd without pauses in:

	      However a key
	      can be reserved in exclusive mode and on a per screen basis
	      with <command>key_add</command> just fine. (In this case
	      the per screen reservation wins.)

One suggestion:

	      However a key
	      can be reserved in exclusive mode and on a per screen basis
	      (with <command>key_add</command>) just fine. In this case
	      the per screen reservation wins.

This shifts the focus to "what" and makes the "how" an aside.

But these are very minor things.  As given, they still are fully understandable.

docs/lcdproc-dev/language.docbook
```diff
@@ -760,14 +803,22 @@
...
 	      Key(s) reserved in exclusive mode will be returned regardless of
 	      which screen is active. They can be used to trigger a special
 	      feature or to make a screen come to foreground.
+	    </para>
+	    <para>
 	      Note that you cannot reserve a key in exclusive mode when an
-	      other client has reserved it in shared mode.
+	      other client has reserved it in shared mode. However a key
+	      can be reserved in exclusive mode and on a per screen basis
+	      with <command>key_add</command> just fine. (In this case
```
ethandicks, line 820: Minor edit here to move the parentheses from "In... wins." to around "with... fine."

docs/lcdproc-dev/language.docbook
```diff
@@ -401,6 +401,49 @@
...
+	    <command>key_del
+	      <option><replaceable>screen_id</replaceable></option>
+	      <option><replaceable>key</replaceable></option>...
+	    </command>
+	    since protocol version 0.4
+	  </term>
+	  <listitem>
+	    <para>
+	      Lift reservation of key(s) by <command>key_add</command>.
+	      Arguments are processed sequentially and for each one either
```
ethandicks, line 440: Minor edit here to say "sequentially, and"... "generated, or..." and "message if the key".

ethandicks (approved):

haraldg:
Thanks for the feedback. This has been very helpful.

PR #124, opened 2019-06-06
G15: Add support for directly accessing the LCD

jwrdegoede:
Add a helper library for accessing /dev/hidraw# devices. The main purpose of this helper-lib is to allow drivers to specify a list of supported products and then have the lib figure out which /dev/hidraw# to open;

And add support to the g15 driver for it directly accessing the LCD through the
/dev/hidraw interface.

There are several reasons to do this:

1. LCDd itself is a daemon for driving LCD screens, it really should not
depend on yet another daemon for something which it can easily do itself.

2. g15daemon detaches the kernel driver from the USB interface it used,
this causes several problems:

Problem 1. On the G510 the same USB interface is used for both the media keys (play/pause prev/next song, stop, volume up/down/mute) as well as for the LCD panel. This means that g15daemon causes the media keys on the G510 model to stop working.

Problem 2. The extra macro keys on these keyboards are really just extra keys, all the macro stuff is done in software on the host. There are kernel patches pending upstream which move the handling of these extra keys to the kernel, where it belongs. These patches make the kernel to send KEY_MACRO# events for these keys. This is part of an upstream effort to create a single, generic, vendor agnostic solution for keyboards with macro-keys, see: https://github.com/libratbag/libratbag/issues/172  g15daemon conflicts with this, esp. the unbinding of the kernel driver it does.


server/drivers/g15.c
```diff
@@ -75,41 +88,45 @@ MODULE_EXPORT int g15_init (Driver *drvthis)
 	p = (PrivateData *) calloc(1, sizeof(PrivateData));
 	if (p == NULL)
 		return -1;
-	if (drvthis->store_private_ptr(drvthis, p))
+
+	p->g15screen_fd = -1;
+
+	if (drvthis->store_private_ptr(drvthis, p)) {
```
haraldg: store_private_ptr() is one of the foolish aspects of our API. There is nothing wrong with just assigning the pointer directly.

Even if you insist on using the function, it always returns 0, so the check is just unnecessary.

server/drivers/g15.c
```diff
@@ -75,41 +88,45 @@ MODULE_EXPORT int g15_init (Driver *drvthis)
...
 	if((p->g15screen_fd = new_g15_screen(G15_G15RBUF)) < 0)
 	{
 		report(RPT_ERR, "%s: Sorry, cant connect to the G15daemon", drvthis->name);
+		g15_destroy(p);
 		return -1;
 	}
 
 	/* make sure the canvas is there... */
 	p->canvas = (g15canvas *) malloc(sizeof(g15canvas));
 	if (p->canvas == NULL) {
```
haraldg: I would treat malloc of small data structures just as fail safe.

However in this case, since it's a fixed size allocation, I suggest to just make it a member of private data directly.

server/drivers/g15.c
```diff
@@ -65,6 +65,19 @@ MODULE_EXPORT int stay_in_foreground = 0;
 MODULE_EXPORT int supports_multiple = 0;
 MODULE_EXPORT char *symbol_prefix = "g15_";
 
+// Destroy a G15 PrivateData struct, which may have been partially initialized
+//
+static void g15_destroy(PrivateData *p)
```
haraldg: Why not use g15_close() instead of the helper?

server/drivers/hidraw_lib.c
```diff
@@ -0,0 +1,155 @@
...
+void lib_hidraw_send_output_report(struct lib_hidraw_handle *handle,
+				   unsigned char *data, int count)
+{
+	int result;
+
+	if (handle->fd != -1) {
+		result = write(handle->fd, data, count);
+		/* Device unplugged / lost connection ? */
+		if (result == -1 && errno == ENODEV) {
+			fprintf(stderr, "Lost hidraw device connection\n");
```
haraldg: drivers are not supposed to write to stderr directly.

You might use report() or just be silent.

server/drivers/Makefile.am
```diff
@@ -24,7 +24,7 @@ AM_LDFLAGS = @LDSHARED@
 lcdexecbindir = $(pkglibdir)
 lcdexecbin_PROGRAMS = @DRIVERS@
 EXTRA_PROGRAMS = bayrad CFontz CFontzPacket curses CwLnx debug ea65 EyeboxOne futaba g15 glcd glcdlib glk hd44780 i2500vfd icp_a106 imon imonlcd IOWarrior irman irtrans joy lb216 lcdm001 lcterm linux_input lirc lis MD8800 mdm166a ms6931 mtc_s16209x MtxOrb mx5000 NoritakeVFD Olimex_MOD_LCD1x9 picolcd pyramid rawserial sdeclcd sed1330 sed1520 serialPOS serialVFD shuttleVFD sli stv5730 SureElec svga t6963 text tyan ula200 vlsys_m428 xosd yard2LCD
-noinst_LIBRARIES = libLCD.a libbignum.a
+noinst_LIBRARIES = libLCD.a libbignum.a libhidraw.a
```
haraldg: Since this will be statically linked, why bother having a library in the first place? In principle I don't care either way, in practice I'm concerned that this will cause build errors on non-linux OSes.

LCDd.conf
```diff
@@ -356,6 +356,9 @@ keypad_test_mode=no
 
 ## g15 driver for Logitech G15 Keyboard LCDs ##
 [g15]
+# Set this to yes to directly access the keyboard through a /dev/hidraw#
+# device instead of connecting to g15daemon
+#HidrawAccess=no
```
haraldg: Is there any way to have this auto-detected? E.g. is it save to assume that if we can connect to the g15daemon then we have to use it, because it did detach the kernel driver?

haraldg (changes requested):
A couple of comments inline. More generally:

* Is libg15daemon_client still needed in your new code path? If not, maybe it does make sense to change the build system to make it optional?

* You should updatet the g15 section in the userguide to reflect your changes.

server/drivers/g15.c
jwrdegoede: Embedding the structs into the private data struct is a good idea, will fix.

server/drivers/g15.c
jwrdegoede: I was just moving the existing code around, but I agree that directly setting the pointer is much better, so  since I'm touching the code anyways I will change it to use a direct assignment for the pointer in the next version.

server/drivers/g15.c
jwrdegoede: Good point, will fix.

server/drivers/hidraw_lib.c
jwrdegoede: Ok, will switch to report.

server/drivers/Makefile.am
jwrdegoede: The hidraw code was deliberately writtent to be generic, so putting it in a lib seems like the right thing to do to me.

haraldg: Sorry, I was being unclear: Putting this into its own file of course is the right thing to do. But it need not be a lib, it could be just an extra source file like we do it with i2c.c

I'm not opposed to doing it as lib, but then you need to make sure that it doesn't break builds on non-linux OSes, where presumably the hidraw headers are not present.

haraldg:
I cherry-picked eebac56 for now.

server/drivers/Makefile.am
jwrdegoede: A good point (non Linux platforms not having the linux/hidraw.h header. I will do as you suggest for the next version, dropping the libhidraw.a and instead add hidraw_lib.c to g15_SOURCES.

LCDd.conf
jwrdegoede: That is a good idea, this also combines nicely with making libg15daemon_client optional now that we also support direct access. So I've implemented both.

jwrdegoede:
I've just pushed a new version addressing all your comments.

haraldg:
I have merged the commits from this PR on the shell. Thus closing manually.

jwrdegoede:
> I have merged the commits from this PR on the shell. Thus closing manually.

Great, thank you.

PR #126, opened 2019-06-09
Various improvements for the Linux input driver

jwrdegoede:
Various improvements for the Linux input driver:

- Allow opening the input device by device-name
- Re-acquire the device on connection loss
- Drop no longer needed custom mx5000 input handling

### DIFF
--- a/server/drivers/linux_input.c
+++ b/server/drivers/linux_input.c
@@ -42,11 +42,11 @@ struct keycode {
 static struct keycode *
 keycode_create(const char *configvalue)
 {
-	int code;
+	long code;
 	char *button;
 	struct keycode *ret;
 
-	code = atoi(configvalue);
+	code = strtol(configvalue, NULL, 0);
 	if (code < 0 || code > UINT16_MAX)
 		return NULL;
 
--- a/server/drivers/linux_input.c
+++ b/server/drivers/linux_input.c
@@ -69,6 +69,8 @@ keycode_create(const char *configvalue)
 /** private data for the linux event device driver */
 typedef struct linuxInput_private_data {
 	int fd;
+	/* For re-acquiring the device on connection loss when openen by name */
+	const char *name;
 	LinkedList *buttonmap;
 } PrivateData;
 
@@ -183,6 +185,7 @@ linuxInput_init (Driver *drvthis)
 					drvthis->name, s);
 			return -1;
 		}
+		p->name = s;
 	}
 
 	for (i = 0; (s = drvthis->config_get_string(drvthis->name, "key", i, NULL)) != NULL; i++) {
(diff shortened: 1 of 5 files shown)


jwrdegoede:
I've just pushed one more fix: "Only use the default mappings when there is no user config"

server/drivers/linux_input.c
```diff
@@ -183,6 +185,7 @@ linuxInput_init (Driver *drvthis)
 					drvthis->name, s);
 			return -1;
 		}
+		p->name = s;
```
haraldg, line 188: I believe we should strdup() here. It probably works without, but I think we have no guarantee, that the server keeps the memory s is pointing at around.

Otherwise this PR looks perfect to me.

jwrdegoede:
I've just pushed one more fix: "server/drivers/linux_input: Consume all events until we find a valid keypress"

server/drivers/linux_input.c
jwrdegoede: Heh, I actually checked the lifetime of the string returned by config_get_string before doing things like this, the string is valid until config_clear() gets called.

The only caller of config_clear() is do_reload() from server/main.c which first calls drivers_unload_all() so the driver gets closed before the string returned from config_get_string() becomes invalid. IOW no need to strdup.

haraldg: Yes, it works. But the point is: We have an API between server and drivers and ATM this API doesn't guarantee that the server will keep the string around. Of course we can just update the API in this regard. (I kind of would love to - I hate the amount of pointless strdup() we do for this reason.) So the real question is: Should we do this? I kind of understand the point between not mixing server allocated memory and driver allocated memory in the same data structure, so I never dared proposing something like that. But I'm certainly willing to be convinced.

However, actually the whole question isn't that important, because we probably will switch our configuration management soon after the next release to libelektra ...

server/drivers/linux_input.c
```diff
@@ -288,20 +286,35 @@ linuxInput_get_key (Driver *drvthis)
...
+
+/**
+ * Helper function to convert a key code to a key name.
+ * \param p      Pointer to driver linuxInput PrivateData structure
+ * \param code   Linux KEY_ key-code to convert
+ * \retval       String representation of the key;
+ *               \c NULL for nothing available / error.
+ */
+const char *
+linuxInput_key_code_to_key_name (PrivateData *p, uint16_t code)
```
haraldg, line 304: Make this function static?

server/drivers/linux_input.c
jwrdegoede: Ok, if this is going to change anyways I'm going to keep this as is. Let me know if I misunderstood and you still want me to add the strdup.

server/drivers/linux_input.c
jwrdegoede: Ack, good one, I will push a new version with this fixed in a minute.

jwrdegoede:
As promised I've just pushed a new version with the missing static added.

PR #127, opened 2019-06-23
Guard alloca.h

fmertz:
Linux defines **alloca** in alloca.h, but FreeBSD defines **alloca** in stdlib.h

Added a build-time AC_CHECK_HEADERS for alloca.h, which drives HAVE_ALLOCA_H in config.h


haraldg:
Seems like I have messed up. I was believing that alloca() is C89, but that's not actually true. I now have switched the code to C99 syntax instead, which should be more portable. Can you test if #128 solves the problem for you?

fmertz:
#128 compiles just fine under FreeBSD.

haraldg:
Ok,  merged then other PR then.

PR #128, opened 2019-06-24
server: Don't use alloca() - use C99 instead

haraldg:
Turns out alloca() is declared in different header files on GNU and BSD
systems. Therefore we switch from alloca() to C99, which should be
easier portable.

Signed-off-by: Harald Geyer <harald@ccbib.org>


fmertz:
Great fix, compiles just fine under FreeBSD.

PR #130, opened 2019-07-01
Remove obsolete ftime()

fmertz:
FreeBSD 11.2 warns of the use of the obsolete ftime() function when compiling the ICP driver. A run-time error occurs when the driver is loaded as ftime() is unresolved (libcompat is not installed by default)

Error was reported on the pfSense Forum: [Problem with A125 LCD board with LCDProc](https://forum.netgate.com/topic/144031/problem-with-a125-lcd-board-with-lcdproc)

As the documentation suggests, replaced with a call to gettimeofday()


haraldg:
cherry-picked the relevant commit on the comand line.

PR #143, opened 2019-12-28
gethostbyname() is deprecated

emteejay:
Replace gethostbyname() which is obsolete with getaddrinfo().


haraldg:
Thanks for looking into this. I have cherry-picked two commits, that looked good. However I haven't merged the actually important fix: I think changing the port setting from int to string might break upgrades (and is complicated anyway). Instead I suggest to convert the port to a string just before calling getaddrinfo().

Since I merged some commits, you will have to rebase your branch.

emteejay:
The change of the port from int to string is private to the driver, the
configuration file is just text, the drivers are free to interpret it as
they wish.

haraldg:
I guess they are and it won't cause issues on upgrades and we probably
don't care about downgrades. But if we really are to allow service
names in addition to port numbers, we should at least update the
documentation to reflect this.

emteejay:
I recommend leaving the documentation as it is for now. I think it is highly improbable that the LIRC team will ever register the UDP port that they use.

emteejay:
By the way I have only been able to test using IPv4, The LIRC udp driver has not been updated to support IPv6.

haraldg:
Very well, I still don't like the mess we introduce with this, but no point in arguing over obscure legacy code. I have cherry-picked the commit and close this PR.

PR #147, opened 2020-01-29
Longannano

Martoni:
This is a driver proposal for the Longan Nano made by Sipeed.
Longan Nano is a little USB board including a GD32VF microcontroller (with riscv bumblebee core) and a RGB amoled screen.

The driver is based on rawserial with some modification. It is related to firmware code available [here](https://github.com/Martoni/LnLcdProc).


haraldg:
Hi, thanks for working on this. Please send a clean PR - that's easier to review. Also you will need to add documentation and drop any local changes to LCDd.conf.

But since you also write the firmware for the device, I wonder whether a new driver is necessary at all. Can't you implement a protocol on the firmware side, that is compatible with one of the existing drivers?

Martoni:
> Hi, thanks for working on this. Please send a clean PR - that's easier to review. Also you will need to add documentation and drop any local changes to LCDd.conf.
> 


Ok I will make some cleaning to my fork.

> But since you also write the firmware for the device, I wonder whether a new driver is necessary at all. Can't you implement a protocol on the firmware side, that is compatible with one of the existing drivers?

Good question, I think it's possible with rawserial yes. Will see that.

haraldg:
Well over a year without activity and the last suggestion was to not have a new driver at all, so I'm closing this.

Please submit a new PR if the topic needs further discussion.

PR #151, opened 2020-04-03
fortune.pl - scale fortunes for LCD size

ethandicks:
Updated fortune.pl to work with different sized panels than 20x4 by using reported LCD size from server connect response.  Also added verbosity argument and diagnostic messages


haraldg (approved):
Looks good to me. Any reason you wait with merging this?

ethandicks:
I tested it on a couple of screen sizes but I wanted feedback from anyone else that had a large screen.  That's it.

PR #157, opened 2020-07-24
Fix broken behavior of checkbox/ring widgets in wizard menus

### LINKED ISSUE #156: Questionable behavior of checkbox and ring widgets in wizards
The LCDproc developers guide explicitly allows `menu_set_item -next` for checkbox and ring widgets. Pressing the `ENTER` on such widgets, surprisingly, has *two* effects:
1. the `value` is updated (checkbox is toggled resp. next ring value selected)
2. the successor item is activated

So jumping through such a wizard using the `ENTER` key will silently change the value of *every* checkbox and ring item, which is probably not what most users expect.

Since both widgets allow changing the value using the `RIGHT` and `LEFT` key as well, this could be easily fixed by just activating the successor item (if present), and leave the value alone. The problem with this approach is that such wizards can't be operated with only 3 keys any more then.

mklein-de:
If a checkbox or ring widget has a successor item set (-next), pressing
the ENTER key should not change its value.

Closes: #156

### DIFF
--- a/docs/lcdproc-user/configuration.docbook
+++ b/docs/lcdproc-user/configuration.docbook
@@ -684,14 +684,18 @@ You can configure what keys the menu should use.
 </variablelist>
 
 <para>
-The minimal keys required for the menu work correctly are the <command>MenuKey</command>,
-the <command>EnterKey</command> and one of <command>UpKey</command> or
-<command>DownKey</command>.
-With these 3 keys the menus can be operated.
+The minimal keys required for most menus to work correctly are the
+<command>MenuKey</command>, the <command>EnterKey</command> and one of
+<command>UpKey</command> or <command>DownKey</command>.
+With these 3 keys the menus can usually be operated.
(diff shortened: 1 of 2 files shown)


haraldg:
I'm inclined to merge this. But waiting two weeks for @willfe @ethandicks @marschap or really everybody to chime in.

ethandicks:
On Fri, Jul 24, 2020 at 3:30 PM haraldg <notifications@github.com> wrote:
> I'm inclined to merge this. But waiting two weeks for @willfe @ethandicks @marschap or really everybody to chime in.

I'll take a look as soon as I can.

haraldg:
@ethandicks it's been much more then two weeks. Do you still intend to review this or should I go ahead an merge?

ethandicks (approved):

ethandicks:
I did not compile and test the changed behavior, but reading through the code, the fix appears reasonable.

PR #158, opened 2020-07-25
ugly "fix" to prevent "buffer overflow detected" errors when the LCD …

ghost:
…width is bigger than 16 and device-name as well

### DIFF
--- a/clients/lcdproc/disk.c
+++ b/clients/lcdproc/disk.c
@@ -59,7 +59,7 @@ disk_screen(int rep, int display, int *flags_ptr)
 
 	/* Holds info to display (avoid recalculating it) */
 	struct disp {
-		char dev[16];
+		char dev[256];
 		char cap[8];
 		int full;
 	} table[256];


clients/lcdproc/disk.c
```diff
@@ -59,7 +59,7 @@ disk_screen(int rep, int display, int *flags_ptr)
 
 	/* Holds info to display (avoid recalculating it) */
 	struct disp {
```
haraldg, line 61: This function seems broken in many ways. Most notably table is not static, so it is recalculated anyway.

Maybe just merge the two for loops together and get rid of table completely?

haraldg:
@vintagepc Maybe this PR addresses the crash you are seeing?

vintagepc:
Ooh, that could definitely be it, the disk definitely has a path longer than 16 chars and the display is a 40x4

haraldg:
Fixed by #181

PR #161, opened 2020-09-10
menu_goto: allow special menu id value "_quit_"

mklein-de:
this makes it possible to close the client menu programatically


haraldg:
I'm inclined to merge this. But since this is technically a protocol change waiting two weeks for @willfe @ethandicks @marschap or really everybody to chime in.

Also we should note in the documentation, that this is only available in the new protocol version. (Good timing, that we are incrementing protocol version with 0.5.10 anyway.)

ethandicks:
This seems a trivial addition to the protocol if we are bumping it already.  I see no problems with incorporating it.  Definitely need a documentation update to go with.

PR #163, opened 2021-01-10
Added "BCM2711" of Raspberry Pi 400 and his changed gpio_base_address in a if...else Ladder.

Corbomit:
Greetings, I'm a big Fan of LCDproc, but realized it does not work on the new Raspberry Pi 400, so I looked into it and found the new Typename, Revision and the changed GPIO Base Adress.

I hope it's OK to use the absolute value of 19 for the Revision, but I did not found a range.

Thank you for maintaining this and all the best :-)

### DIFF
--- a/server/drivers/hd44780-rpi.c
+++ b/server/drivers/hd44780-rpi.c
@@ -166,6 +166,7 @@ check_board_rev(Driver *drvthis)
 	/* On boards that have been overvolted, the MSB will be set */
 	if ((strcmp(hw, "BCM2708") != 0 &&
 	     strcmp(hw, "BCM2709") != 0 &&
+	     strcmp(hw, "BCM2711") != 0 &&
 	     strcmp(hw, "BCM2835") != 0) ||
 	    rev & 0xFFFFFF == 0) {
 		report(RPT_ERR, "check_board_rev: This board is not recognized as a Raspberry Pi!");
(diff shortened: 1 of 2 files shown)


kripton:
Doesn't the RPI4 also use the BCM2711? Just out of interest: Are you running in 32 or 64 bit mode?

Corbomit:
On the Raspberry Pi 4 it shows
```
more /proc/cpuinfo
model name	: ARMv7 Processor rev 3 (v7l)
...
Hardware	: BCM2835
Revision	: c03111
Model		: Raspberry Pi 4 Model B Rev 1.1
```
Maybe they changed it in other Revisions.

I run them all with 32 bit.

ethandicks:
I have a Raspberry Pi 400 and am going through things now.  I can definitely confirm the hardware string ("BCM2711") and that the latest version of LCDproc reports "check_board_rev: This board is not recognized as a Raspberry Pi!" when starting LCDd with the HD44780 driver and 'ConnectionType=raspberrypi'

I will give this patch a try (which I expect will work just fine) but I cannot yet test running an LCD until I finish building a cable harness for 40-pin GPIO ports (my existing harness only fits ancient 26-pin GPIO ports).  Once built, I expect _that_ to work just fine.

I'm happy to merge this once I've seen it run.

ethandicks:
OK.  With this patch, running at reporting level 4 or higher, I now see "check_board_rev: Raspberry Pi 400 detected".  I'll finish my harness with the default GPIO pin mapping and test the hardware.

Looking good so far.

ethandicks:
OK.  It took 2-3 tries to find a bare HD44780 module with a positive bias voltage but I was finally able to verify this end-to-end.

PR #167, opened 2021-02-09
Update Code of Merit URL

WardBenjamin:
The old http://code-of-merit.org page now redirects to https://www.mayan-edms.com/

From the new website:

>This website is independent and maintained by Valentino Giudice.
>
> The Code of Merit used to be mantained by its original author but he discontinued every one of his projects except for Mayan EDMS. This website is an attempt to keep the Code of Merit available.


willfe:
Looks good to me.

PR #172, opened 2021-05-03
lcdexec: fix bug that prevented command with arg

cuanlategan:
It looks like this bug slipped through the cracks with commit c39c4f14

### DIFF
--- a/clients/lcdexec/lcdexec.c
+++ b/clients/lcdexec/lcdexec.c
@@ -418,10 +418,10 @@ static int process_response(char *str)
 			 * - command entry without args
 			 * - last arg of a command entry with args */
 			if (((entry->type == MT_EXEC) && (entry->children == NULL)) ||
-			    ((entry->type & MT_ARGUMENT) && (entry->next == NULL))) {
+			    ((entry->type == MT_AUTOMATIC) && (entry->next == NULL))) {
 
 				// last arg => get parent entry
-				if ((entry->type & MT_ARGUMENT) && (entry->next == NULL))
+				if (entry->type == MT_AUTOMATIC)
 					entry = entry->parent;
 
 				if (entry->type == MT_EXEC)
--- a/clients/lcdexec/menu.c
+++ b/clients/lcdexec/menu.c
@@ -234,7 +234,7 @@ MenuEntry *menu_read(MenuEntry *parent, const char *name)
 			me->next = NULL;
 			me->children = NULL;
 			me->numChildren = 0;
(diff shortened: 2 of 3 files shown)


haraldg (changes requested):
Hi, thanks for looking into this! It seems the definition of MT_ACTION was rash and not tested well enough.

However I think your proposal is not a clean fix. Not sure if it actually breaks some functionality, but it breaks the logic of flags enough, that just trying to figure out, whether the fix is okay is a pain.

I suggest reverting my commit and replacing it by a new clean solution to the original problem.

cuanlategan:
Hi, thanks for the quick response.

Just so we are on the same page, your suggestion is to revert commit c39c4f1? 

From what I can tell, when MT_AUTOMATIC was introduced it made the MT_ARGUMENT mask redundant

Is this [branch](https://github.com/cuanlategan/lcdproc/tree/lcdexec_fix2) more like what you are after?

haraldg:
On 04.05.2021 14:08, cuanlategan wrote:
> Just so we are on the same page, your suggestion is to revert commit
> c39c4f1?

Yes, it seems I introduced a subtle bug while fixing an obvious problem
because I changed the code without actually comprehending it. My
thinking is, that by going back to its original state, it is easier to
comprehend and actually improve.

> From what I can tell, when MT_AUTOMATIC was introduced it made the
> MT_ARGUMENT mask redundant

I doubt it.

> Is this
> [branch](https://github.com/cuanlategan/lcdproc/tree/lcdexec_fix2)
> more like what you are after?

I'd guess your change breaks the non-automatic menus, doesn't it?

cuanlategan:
> I'd guess your change breaks the non-automatic menus, doesn't it?

not from what I can tell. 

Shouldn't a non-automatic menu command have an entry type of MT_EXEC?

https://user-images.githubusercontent.com/11440243/117061216-34b98980-ad22-11eb-9b69-d90f5faa6cdd.mp4

haraldg:
On 04.05.2021 20:49, cuanlategan wrote:
>> I'd guess your change breaks the non-automatic menus, doesn't it?
> not from what I can tell
>
> 
> https://user-images.githubusercontent.com/11440243/117061216-34b98980-ad22-11eb-9b69-d90f5faa6cdd.mp4

Thanks for sharing your test. I guess I'll need to actually
comprehend the original code to feel confident merging this.

Please ping me again if you don't hear back from me for a
while.

cuanlategan:
closing as an alternative solution #173 fixes this problem

PR #173, opened 2021-05-06
lcdexec: fix bug that prevented command with arg

haraldg:
Revert two commits, that changed the code without actually comprehending it.

This is an alternative fix to #172.


haraldg:
@cuanlategan does this address all your conerns?

cuanlategan:
> @cuanlategan does this address all your conerns?

Seems to work well, I'll close #172

PR #180, opened 2021-07-25
Add Disk ignore list, fix buffer overflow.

### LINKED ISSUE #178: Disk screen has buffer overflow and crashes LCDproc
Observed today when I plug in and mount a USB drive on my computer, LCDProc will crash and refuse to re-launch. 

v0.5.7

```
 lcdproc -f -c /etc/lcdproc.conf 
*** buffer overflow detected ***: terminated
Aborted (core dumped)

```

Disabling the disk screen in /etc/lcdproc.conf (or unmounting the drive and restarting lcdproc) will resolve the issue. 
I have two drives that do this and several others that do not. The two that cause it are both USB3.0/32GB drives formatted as FAT32. 

Problem does not seem USB3 related as it also manifests if the same drive is plugged into a USB2 port. 

Bottom-most drive is the culprit. 
image

Display is a USB-serial connected 40x4 using the MtxOrb protocol.

### LINKED ISSUE #179: [Feature] Disk ignore list
Just adding an issue here so it's identified/searchable and there's a place for discussion. 

I'm working on the ability to configure a list of mount points that should be ignored on the Disk screen. 

Modern systems often have "/boot/efi" which is a normal FS (and therefore not excluded by the machine filtering but these (or potentially other user-) mountpoints where the disk consumption stats are less valuable than the display line they occupy :smile: 

Proposal:
I plan to add a config KVP option to the [Disk] section of the config file that allows you to list up to 10 mountpoints that should not be shown on the Disk screen. I do not currently have plans to support wildcards but will certainly consider it if there's a strong argument in favor of. 

I'll also include #178 and #158 since this requires reworking that particular section of code anyway.

vintagepc:
- Fix the buffer overflow for wide displays with long mount paths.
  -  Closes #178 
  -  Supersedes #158 
- Add an "ignore" list feature to the lcdproc config file for non-interesting mount points. 
  - Closes #179

Up to 10 mountpoints can be ignored, but this limit can be adjusted by changing the appropriate #define and recompiling.

### DIFF
--- a/clients/lcdproc/disk.c
+++ b/clients/lcdproc/disk.c
@@ -22,6 +22,8 @@
 #include "config.h"
 #endif
 
+#include "shared/configfile.h"
+#include "shared/report.h"
 #include "shared/sockets.h"
 
 #include "main.h"
@@ -30,6 +32,37 @@
 #include "disk.h"
 #include "util.h"
 
+#define DISK_IGNORE_MAX 10
+
+char* disk_ignore[DISK_IGNORE_MAX];
+
+int ignore_count = 0;
+
+/** Parses the "Ignore" line in the config to let you hide
+ * uninteresting "normal" filesystems (like /boot/efi)
+ *  * \return  0 on success, -1 if malloc failure
+ */
+
+static int
+disk_process_configfile(void)
+{
+	debug(RPT_DEBUG, "%s()", __FUNCTION__);
+
+	/* Read config settings */
+	int count = 0;
+	do {
+		disk_ignore[count] = strdup(config_get_string("Disk", "Ignore", count, ""));
+		if (disk_ignore[count] == NULL) {
+			report(RPT_CRIT, "malloc failure");
+			return -1;
+		}
+		if (*disk_ignore[count] == '\0')
+			break;
+		count++;
+	} while (count<DISK_IGNORE_MAX);
+	ignore_count = count;
+	return 0;
+}
 
 /**
  * Gives disk stats.
@@ -57,21 +90,16 @@ disk_screen(int rep, int display, int *flags_ptr)
 	mounts_type mnt[256];
 	int count = 0;
 
-	/* Holds info to display (avoid recalculating it) */
-	struct disp {
-		char dev[16];
-		char cap[8];
-		int full;
-	} table[256];
 	int i;
 	static int num_disks = 0;
 	static int dev_wid = 6;
 	static int gauge_wid = 6, gauge_scale, hbar_pos;
-
-	u_int64_t size;
+	int ig_offset = 0;
 
 	if ((*flags_ptr & INITIALIZED) == 0) {
 		*flags_ptr |= INITIALIZED;
+
+		disk_process_configfile();
 
 		dev_wid = (lcd_wid >= 20) ? (lcd_wid - 8) / 2 : (lcd_wid / 2) - 1;
 		hbar_pos = (lcd_wid >= 20) ? dev_wid + 10 : dev_wid + 3;
@@ -101,21 +129,6 @@ disk_screen(int rep, int display, int *flags_ptr)
 	/* Fill the display structure... */
 	sock_send_string(sock, "widget_set D err1 0 0 .\n");
 	sock_send_string(sock, "widget_set D err2 0 0 .\n");
-	for (i = 0; i < count; i++) {
-		if (strlen(mnt[i].mpoint) > dev_wid)
-			sprintf(table[i].dev, "-%s", (mnt[i].mpoint) + (strlen(mnt[i].mpoint) - (dev_wid - 1)));
-		else
-			sprintf(table[i].dev, "%s", mnt[i].mpoint);
-
-		table[i].full = !mnt[i].blocks ? gauge_scale :
-			gauge_scale * (u_int64_t) (mnt[i].blocks - mnt[i].bfree)
-			/ mnt[i].blocks;
-
-		size = (u_int64_t) mnt[i].bsize * mnt[i].blocks;
-		memset(table[i].cap, '\0', 8);
-
-		sprintf_memory(table[i].cap, (double) size, 1);
-	}
 
 	/*
 	 * Display stuff...  (show for two seconds, then scroll once per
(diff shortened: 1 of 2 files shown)


clients/lcdproc/disk.c
```diff
@@ -30,6 +32,37 @@
 #include "disk.h"
 #include "util.h"
 
+#define DISK_IGNORE_MAX 10
+
+char* disk_ignore[DISK_IGNORE_MAX];
+
```
haraldg, line 38: We also have a linked list implementation in shared/LL.h, which would allow to have unlimited entries. However this is up to you, I'm fine either way.

clients/lcdproc/disk.c
```diff
@@ -30,6 +32,37 @@
...
+{
+	debug(RPT_DEBUG, "%s()", __FUNCTION__);
+
+	/* Read config settings */
+	int count = 0;
+	do {
+		disk_ignore[count] = strdup(config_get_string("Disk", "Ignore", count, ""));
+		if (disk_ignore[count] == NULL) {
+			report(RPT_CRIT, "malloc failure");
+			return -1;
```
haraldg, line 57: I wouldn't bother about malloc() failing. On modern systems that doesn't happen unless we are in really big trouble. (Usually the OOM-killer would kill us, before strdup() even returns. OTOH the error path is awkward: Should we ever take it, we don't update `ignore_count` even though some values might have been obtained successfully. In short: Lots of complication and confusion for very little gain.

strdup()ing an emptry string is a bit wasteful.

I think using `NULL` as default and checking before strdup()ing would make everything more straightforward.

clients/lcdproc/disk.c
```diff
@@ -57,22 +90,17 @@ disk_screen(int rep, int display, int *flags_ptr)
...
 	static int gauge_wid = 6, gauge_scale, hbar_pos;
-
-	u_int64_t size;
+	int ig_offset = 0;
 
 	if ((*flags_ptr & INITIALIZED) == 0) {
 		*flags_ptr |= INITIALIZED;
 
+		disk_process_configfile();
+
```
haraldg, line 103: If you inlined the code instead of the call here, you could have `disk_ignore` as local static instead of global. Again, I don't really care either way.

clients/lcdproc/disk.c
```diff
@@ -124,32 +137,71 @@ disk_screen(int rep, int display, int *flags_ptr)
 	sock_printf(sock, "widget_set D f 1 2 %i %i %i %i v 12\n", lcd_wid, lcd_hgt, lcd_wid, count);
 	for (i = 0; i < count; i++) {
 		char tmp[lcd_wid + 1];	/* should be large enough */
+		char cap[8] = {'\0'};
+		char dev[256];
```
haraldg, line 141: Can we use dev_wid here instead of the magic number 256?

clients/lcdproc/disk.c
```diff
@@ -124,32 +137,71 @@ disk_screen(int rep, int display, int *flags_ptr)
...
+		u_int64_t size = 0;
+		int full = 0;
 
-		if (table[i].dev[0] == '\0')
+		// Eewww, goto... but shy of restructuring everything it's 
+		// the cleanest way to skip the rest of the outer loop. :-/
+		next_mnt:
+
```
haraldg, line 148: I'm kind of a fan of goto and use it regularly. But jumping backwards in the program always is more confusing then other uses.

clients/lcdproc/disk.c
```diff
@@ -124,32 +137,71 @@ disk_screen(int rep, int display, int *flags_ptr)
...
-		if (i >= num_disks) {	/* Make sure we have enough lines... */
-			sock_printf(sock, "widget_add D s%i string -in f\n", i);
-			sock_printf(sock, "widget_add D h%i hbar -in f\n", i);
+		// Prescreen for entries to ignore.
+		for (int ign_ct=0; ign_ct<ignore_count; ign_ct++) {
+			if (strcmp(mnt[i].mpoint, disk_ignore[ign_ct])==0) {
+				ig_offset++;
+				i++;
+				goto next_mnt; // Behold, the one legitimate use for goto ;-)
+			}
```
haraldg, line 158: There are many legitimate uses for goto, this is not one of them. Manually messing with loop counters is _exactly_, what makes programs hard to understand and why goto is so unpopular.

I suggest you define a helper function and use something like:
```
if (is_ignored(mnt[i].mpoint))
        continue;
```

clients/lcdproc/disk.c
```diff
@@ -124,32 +137,71 @@ disk_screen(int rep, int display, int *flags_ptr)
...
 		}
-		if (lcd_wid >= 20) {	/* 20+x columns */
-			sprintf(tmp, "%-*s %6s E%*sF", dev_wid, table[i].dev, table[i].cap, gauge_wid, "");
+
+		// Parse/format the incoming FS data:
+		int i_widget = i-ig_offset; // Index into the widget array due to "ignore" list
+
```
haraldg, line 163: Just having `i_widget` as secondary counter in the loop probably would be a lot simpler.

clients/lcdproc/disk.c
```diff
@@ -124,32 +137,71 @@ disk_screen(int rep, int display, int *flags_ptr)
 	sock_printf(sock, "widget_set D f 1 2 %i %i %i %i v 12\n", lcd_wid, lcd_hgt, lcd_wid, count);
 	for (i = 0; i < count; i++) {
 		char tmp[lcd_wid + 1];	/* should be large enough */
+		char cap[8] = {'\0'};
```
haraldg, line 140: I'm honestly not good enough with C to know, if this is equivalent to the memset() code we had before?

clients/lcdproc/disk.c
```diff
@@ -124,32 +137,71 @@ disk_screen(int rep, int display, int *flags_ptr)
...
 		sock_printf(sock, "widget_set D h%i %i %i %i\n",
-					i, hbar_pos, i + 1, table[i].full);
+					i_widget, hbar_pos, i_widget + 1, full);
 	}
 
 	/* Now remove extra widgets... */
-	for (; i < num_disks; i++) {
-		sock_printf(sock, "widget_del D s%i\n", i);
-		sock_printf(sock, "widget_del D h%i\n", i);
+	for (int rm = i-ig_offset; rm < num_disks; rm++) {
```
haraldg, line 196: Can't your reuse the `i_widget` counter here? Having multiple variables meaning the same thing just makes everything more difficult to comprehend.

haraldg (changes requested):
Hi! Thanks for looking into this!

Many comments inline, but most of them are just suggestions. However, we definitely need updated documentation (lcdproc-config.5.in and the user guide).

Also if it isn't too much work, please make separate commits for the bugfix and the new feature. That way, things are _much_ easier to review. Also distributions might like to backport the bug fix without also adding new features.

And finally I ask, that you use the same spacing style as is used in the rest of the file.

vintagepc:
Thanks, will review and make the requested changes shortly.

I did think about separate feature/bugfixes  and I can definitely understand the fix-only sentiment in a stability perspective (e.g. minimally invasive changes to reduce risk of breakage) but in this case there isn't really a "lite" version of just the bugfix shy of the array resize in the other open PR; 

If you still prefer I do so then I will split them out (I'm the guest here, after all) but TBH given the scope of the changes it does feel a little like formality for the sake of formality because the bulk of the changes are for the cleanup/bugfix.

haraldg:
As far as "easier to review" is concerned, the big difference is the new variables, that you introduce for the new feature. That makes matching the old code with the new code quite a bit harder.

You are right, that the merging of the two loops as part of the crash fix is quite invasive by itself. But IMO this is just more reason to two the change in to steps, as each one is non-trivial. (And this is even more true, when you add the changes to the user documentation.)

As lcdproc itself is a rather small project, many things might seem like a formality. Still I think it is worth following them as long as they don't cause too much extra work.

clients/lcdproc/disk.c
vintagepc: Yes, brace initializers will zero-fill unspecified elements, e.g. from CPPReference on C array initialization: 

`int z[3] = {0}; // z has type int[3] and holds all zeroes`

haraldg: Just for the benefit of my education: Then `char cap[8] = {};` would work as well?

vintagepc: In C++, yes - but I do not think an empty initializer is allowed in to-spec C (though some compilers might support it anyway)

clients/lcdproc/disk.c
vintagepc: Addressed in #181

clients/lcdproc/disk.c
vintagepc: Valid point; I'll remove that. This was just copy-pasted from `iface.c`'s handling in an attempt to be consistent with established patterns :smile:

clients/lcdproc/disk.c
vintagepc: Sure, I agree. It's small enough there's no readability benefit from segregating it.

clients/lcdproc/disk.c
vintagepc: in this scenario does it make sense to condense the two ifs into one? I'm not sure what situations would cause a null mountpoint to appear in the list (except if you run past `count`... in which case we should already be out of the for loop.)

If it is a possibility then I think it makes sense to also increment ig_offset because otherwise you'd get a blank line in the disk screen.

haraldg: Yeah, sure. My code was just a template, nothing to be used directly.

But I do hope, the code can be simplified, by making `i_widget` a loop variable and getting rid of `if_offset` ...

(I hope I understood your question correctly.)

vintagepc: If it helps clarify, this is what I was thinking of:
```
		if (mnt[i].mpoint[0] == '\0' || disk_is_ignored(mnt[i].mpoint, disk_ignore)) {
			ig_offset++;
			continue;
		}
```

For the other point - you are proposing we make the "widget" index the real loop counter and maintain the offset of "skipped" entries in `mnt[]` instead? Maybe I'm missing something but I don't see a "clean" way to maintain two indices that might diverge (because the ignore entries can appear at any position in `mnt[]`) without either explicitly storing them (or an offset) or getting into things like pointer arithmetic. 

My main question was whether it's ever possible to observe a null mnt.name in practice. But I guess it's moot since for our intent it doesn't matter if it's null or on the ignore list, either way it should not be displayed and we should move on to the next entry.

clients/lcdproc/disk.c
vintagepc: Resolved in #182

clients/lcdproc/disk.c
vintagepc: Resolved in #182

clients/lcdproc/disk.c
vintagepc: Resolved in #182

clients/lcdproc/disk.c
vintagepc: I did manage to remove one of the two new vars :+1:  Don't think we can get away from maintaining an offset though

clients/lcdproc/disk.c
vintagepc: In this version i_widget was out of scope since it was loop-local. I've reworked this.

clients/lcdproc/disk.c
haraldg: I doubt I'm any more familiar with the code then you are. So I don't know either. But I think for the purpose of this PR you can assume (even wrongly), that the original code worked sensibly. So if a null mnt.name doesn't make sense, you a free to pretend, it can't happen. :-) If the original code is broken, we always can (and I actually prefer) to fix it in a separate commit.

In C you certainly can have two distinct loop variables. Something like:
```
for (i = 0, j = 0; i < 10; ++i, ++j)
        if (...)
                ++j;
```
is perfectly valid and actually quite common.

vintagepc: Sure, I know you can have multiple loop variables but I don't think it adds any benefit  here because we don't want to auto-increment i_widget on continue; only if we make it through the end of the loop. Unless you're proposing the logic be completely inverted so that the primary loop controller is the widget index and the index into `mnt[]` is ancillary and bumped on continue.

vintagepc:
Closing because it has been superseded by #182 and #181

clients/lcdproc/disk.c
vintagepc: I did consider something more flexible but I'm always wary of unbounded items if there's a sensible limit that is sufficient for most use cases :smile:

clients/lcdproc/disk.c
haraldg: There are several possible patterns. One is to swap primary and secondary loop variable as you propose. (Probably the most intuitive.) An other is to not auto increment the second variable, but increment it at the end of the loop manually. The worst (but probably still better then having an extra variable) would be to decrement before the continue statement. ;-)

I know, this really is nit picking. However at this time lcdproc is entirely maintained and improved by people who are very much not familiar with the code and just stop by to scratch some itch. Thus I feel it is important, that the code be as readable and comprehensible as possible.

vintagepc: Please have a look at #182. I've eliminated the use of `ig_offset` and the sole additional counter is now `i_widget`;  Unfortunately there is no way to make it entirely scoped to just the loop because its value is required afterwards to remove extraneous line entries and update the static count. (This could probably be integrated into the main FOR as well, but at that point the nesting of control structures starts getting a little onerous. 

The change I have made there sounds like it aligns with the second option you describe above as best as possible. 

I certainly don't take any offense at your feedback here; I would likely do the same for any of my projects that were in "maintenance" mode in order to ensure I would be able to pick things up  easily again if necessary. (in fact, that's why I'm in favor of leaving the loop index alone as much as possible; it makes it easier for someone that has a prior familiarity to pick things up again without having to re-learn some things they might still recall. )

haraldg: Yes, I'm perfectly happy with the way you have handled this in #182. 

In the case of lcdproc there is little hope, that people, who know the code, come back. It's more about making it easy to comprehend for new people.

PR #181, opened 2021-07-25
Fix buffer overflow in Disk screen with  wide displays and long mount paths.

### LINKED ISSUE #178: Disk screen has buffer overflow and crashes LCDproc

vintagepc:
Buffer overflow bugfix with rework to remove the now-unnecessary table/loop and an unused variable.

- Fixes #178 
- Supersedes  #158 
- Supersedes #180

### DIFF
--- a/clients/lcdproc/disk.c
+++ b/clients/lcdproc/disk.c
@@ -111,7 +111,6 @@ disk_screen(int rep, int display, int *flags_ptr)
 
 		// This dev sprintf could be condensed with the line formatting to tmp below
 		// but I'm inclined to leave it like this to keep the printf formatting more legible.
-
 		if (strlen(mnt[i].mpoint) > dev_wid)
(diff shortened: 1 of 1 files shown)


haraldg:
Thanks. Now it only took me 2 minutes to see what's going on instead of a full hour. :-)

vintagepc:
Sorry :wink:

PR #190, opened 2021-12-16
drivers/g15: Always depend on freetype2

haraldg:
This is a workaround for an upstream bug. For details see the following
issue. (Closes: #188)

Signed-off-by: Harald Geyer <harald@ccbib.org>

PR #191, opened 2021-12-18
Use PKG_CHECK_MODULES for checking if freetype2 is installed. 

toams:
Fixes compilation on debian based systems


haraldg:
Does this actually work. If I understood you correctly in #190 we also have to replace FT2_CFLAGS with FREETYPE2_CFLAGS? Or what's the difference between the two?

toams:
It turns out that we don't have to replace FT2_CFLAGS with FREETYPE2_CFLAGS to make it work. Only changing AC_CHECK_FT2 to
PKG_CHECK_MODULES was needed. I tested it multiple times.

PR #200, opened 2022-12-19
fix getopt when reload the server

rubens17rodrigues:


haraldg (approved):
I'm not sure whether reload works at all. Maybe this just isn't supported.

But I guess this change makes the code more consistent and correct anyways.

willfe:
Agreed, not sure reloading actually works (that's a separate issue, #230 now), but this improves things. Merging.

Issue #192, opened 2021-12-26
LCD2USB not working with LCDproc but works with LCD Smartie


whc2001:
I have a 2402 LCD module with LCD2USB installed. I tested it on Windows with LCD Smartie and it works well. However when I connected it to my proxmox box and compiled the latest LCDproc from GitHub repo, it does not work at all. After configuration, the backlight would light up, but there is absolutely nothing showing on screen. I wonder if anyone had a similar problem, am I getting a defective unit that is only meant to be used with LCD Smartie, or it's some compatibility issue with LCDproc? 

(Yes I tried lcd4linux and it displayed something, but seems like there is some problem with the timing and the screen would garble the first few characters. Also lcd4linux is a little too simple on functionality so LCDproc looks more promising.)

Here is the config I used:
```
[Server]
DriverPath = /usr/lib/lcdproc/
Driver=hd44780
Hello="Hello World 111111"
Hello="Hello World 222222"

[hd44780]
ConnectionType=lcd2usb
VendorID=0x0403         # Tried with or without
ProductID=0xc630        # Tried with or without
Contrast=500            # Tried with different values, no display
Brightness=1000         # Tried with different values, brightness control works
Backlight=external
Size=24x2
ExtendedMode=yes        # Tried with yes or no
#DelayMult=4            # Tried with or without, 2 through 4
#DelayBus=yes           # Tried with yes or no
```

And here is what the LCDd outputs when running in foreground with maximum debug level:
```
LCDd version 0.5dev starting
Protocol version 0.4, API version 0.5
Using Configuration File: /etc/LCDd.conf
Set report level to 5, output to stderr
LCDd 0.5dev, LCDproc Protocol 0.4

(copyright notice)

Server running in foreground
Listening for queries on 127.0.0.1:13666
screenlist_init()
driver_load(name="hd44780", filename="/usr/lib/lcdproc/hd44780.so")
HD44780: using ConnectionType: lcd2usb
HD44780: selecting Model: extended
HD44780: backlight: external
hd44780: Using hd44780_default charmap
hd_init_lcd2usb: device with firmware version 2.00 found
screenlist_process()
screenlist_switch(s=[_server_screen])
screenlist_switch: switched to screen [_server_screen]
screenlist_process()
screenlist_process()
screenlist_process()
screenlist_process()
screenlist_process()
screenlist_process()
screenlist_process()
screenlist_process()
screenlist_process()
screenlist_switch(s=[_server_screen])
screenlist_process()
[...]

PR #201, opened 2023-01-26
Add DisableBuffer option for LCD2USB driver

whc2001:
This is a workaround for #192. It adds a boolean option `DisableBuffer` to `hd44780-lcd2usb` that ignores the `tx_buf` of LCD2USB as a workaround for Chinese counterfeit LCD2USB modules.

This is only a rough implementation that tested working on my device. I am not sure if this is up to the code of contribution but I mostly copied the existing source.

### DIFF
--- a/server/drivers/hd44780-lcd2usb.c
+++ b/server/drivers/hd44780-lcd2usb.c
@@ -54,6 +54,16 @@ hd_init_lcd2usb(Driver *drvthis)
 {
 	PrivateData *p = (PrivateData *) drvthis->private_data;
 
+	/* Allocate and store connection specific private data */
+	LCD2USBPrivateData *cp = (LCD2USBPrivateData *) calloc(1, sizeof(LCD2USBPrivateData));
+	if (cp == NULL)
+		return -1;
+
+	/* Read driver specific settings */
+	cp->disablebuffer = drvthis->config_get_bool(drvthis->name, "disablebuffer", 0, 0);
+
+	p->connectionprivatedata = cp;
+
 	struct usb_bus *bus;
 
 	p->hd44780_functions->senddata = lcd2usb_HD44780_senddata;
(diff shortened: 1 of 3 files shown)


ConiKost:
@ethandicks This would be worth to get in the upcoming release.

ethandicks:
This definitely sounds interesting.

It would be very helpful if someone with an affected LCD2USB could test the (upcoming) release candidate.  I have an original LCD2USB but I haven't fired it up in a very long time (plus IIRC I modified mine to support a 40x4 LCD).

ethandicks:
> This is a workaround for #192. It adds a boolean option `DisableBuffer` to `hd44780-lcd2usb` that ignores the `tx_buf` of LCD2USB as a workaround for Chinese counterfeit LCD2USB modules.
> 
> This is only a rough implementation that tested working on my device. I am not sure if this is up to the code of contribution but I mostly copied the existing source.

I get what it's trying to do, but it seems to be rather complex to allocate a private buffer and make a 1-byte entry to remember the value from a config option.  I definitely want to look around to see how it fits and if there's already a pattern for how to handle options like this.

whc2001:
Strange, I received the latest reply by @ethandicks through email but when I clicked view it's not there.

Basically I just copied the whole `PrivateData` thing from another existing driver that I don't remember which. All in all this is just a propose of workaround that's working for me, so feel free to do anything. Thanks a lot for the effort and I'll wait patiently for the upcoming release with my local fork for now.

Magister54:
I'm using this PR with the [LCD2USB chinese clone](https://www.aliexpress.com/item/1005002661543249.html) and it is working flawlessly for 2 weeks now.

ethandicks:
Two items of feedback:

1) Documentation - this new argument needs to be mentioned

2) Why is it using a new privateconnection mechanism for handling an arg instead of the well-used existing drvthis->private_data?

whc2001:
I'm really sorry but I didn't realize there is `drvthis->private_data` when I did this. Again I just copied over from another file, tested it's working and that's it. As for the docs I thought that should be updated only after it's finalized so I didn't change anything. The problem is that I don't have the hardware with me anymore for a while so I am no longer able to test it. In the mean time I am gonna close this for now since after all the credit all goes to [tomwang221812](https://github.com/lcdproc/lcdproc/issues/192#issuecomment-1311268514).

ethandicks:
For a future code submission, docs and code come at same time and get merged at the same time.

The functional part of this looked fine and a resubmission using drvthis->private_data will be favorably reviewed.

whc2001:
Thanks, got it. I'll try again after going back to Canada around next month so I can test with the hardware.

PR #209, opened 2023-11-30
Support seamless HBAR in CFA-635

vintagepc:
The CFA-635 USB display supports seamless characters via 6x8 cells, much like some of the `CFontz` driver displays.

This patch implements support for seamless horizontal bars for this model by setting the appropriate option on `lib_hbar_static` and in the model data for `CFontzPacket.c`

Appearance example/demo/test (MythTV video Playback):

image


ConiKost:
Works great on my CFA-635. @ethandicks Maybe we could add this also for the upcoming release?

PR #216, opened 2024-05-26
Add support for linux /dev/lcd character device

penfold42:
Added linux_devlcd output driver.
    This uses the mainline linux driver which supports hd44780 and maybe other devices
    Some useful resources on the kernel side:
    - https://github.com/torvalds/linux/commit/d47d88361feea2ce11f39bd70467ffc19a61d2d3
    - https://blog.microjoe.org/2019/hd44780-lcd-i2c-screen-using-linux-mainline-charlcd-driver.html
    - https://forums.raspberrypi.com/viewtopic.php?t=285415

### DIFF
--- a/server/drivers/linux_devlcd.c
+++ b/server/drivers/linux_devlcd.c
@@ -1,0 +1,312 @@
+/** \file server/drivers/linux_devlcd.c
+ * LCDd \c driver for linux kernel /dev/lcd device
+ * It displays the LCD screens, one below the other on the terminal,
+ */
+
+/* Copyright (C) 2024 The LCDproc Team
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301
+ */
+
+#ifdef HAVE_CONFIG_H
+#include "config.h"
+#endif
+
+#include <stdlib.h>
+#include <stdio.h>
+#include <string.h>
+#include <errno.h>
+#include <fcntl.h>
+
+#include "lcd.h"
+#include "linux_devlcd.h"
+#include "shared/report.h"
+
+
+/** private data for the \c text driver */
+typedef struct linuxDevLcd_private_data {
+	int width;		/**< display width in characters */
+	int height;		/**< display height in characters */
+	char *framebuf;		/**< fram buffer */
+	FILE* fd;		/**< handle to the device */
+} PrivateData;
+
+
+/* Vars for the server core */
+MODULE_EXPORT char *api_version = API_VERSION;
+MODULE_EXPORT int stay_in_foreground = 0;
+MODULE_EXPORT int supports_multiple = 0;
+MODULE_EXPORT char *symbol_prefix = "linuxDevLcd_";
+
+
+/**
+ * Initialize the driver.
+ * \param drvthis  Pointer to driver structure.
+ * \retval 0       Success.
+ * \retval <0      Error.
+ */
+MODULE_EXPORT int
+linuxDevLcd_init (Driver *drvthis)
+{
+	PrivateData *p;
+	char buf[256];
+	char device[200];
+
+	/* Allocate and store private data */
+	p = (PrivateData *) calloc(1, sizeof(PrivateData));
+	if (p == NULL)
+		return -1;
+	if (drvthis->store_private_ptr(drvthis, p))
+		return -1;
+
+	/* initialize private data */
+
+	/* which device should be used */
+	strncpy(device, drvthis->config_get_string(drvthis->name, "Device", 0, DEFAULT_DEVICE),
+		sizeof(device));
+	device[sizeof(device) - 1] = '\0';
+
+	report(RPT_INFO, "%s: using Device %s",
+	       drvthis->name, device);
+
+	// Set display sizes
+	if ((drvthis->request_display_width() > 0)
+	    && (drvthis->request_display_height() > 0)) {
+		// Use size from primary driver
+		p->width = drvthis->request_display_width();
+		p->height = drvthis->request_display_height();
+	}
+	else {
+		/* Use our own size from config file */
+		strncpy(buf, drvthis->config_get_string(drvthis->name, "Size", 0, TEXTDRV_DEFAULT_SIZE), sizeof(buf));
+		buf[sizeof(buf)-1] = '\0';
+		if ((sscanf(buf , "%dx%d", &p->width, &p->height) != 2)
+		    || (p->width <= 0) || (p->width > LCD_MAX_WIDTH)
+		    || (p->height <= 0) || (p->height > LCD_MAX_HEIGHT)) {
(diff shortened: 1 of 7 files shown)


server/drivers/linux_devlcd.c
```diff
@@ -0,0 +1,312 @@
...
+		return -1;
+	if (drvthis->store_private_ptr(drvthis, p))
+		return -1;
+
+	/* initialize private data */
+
+	/* which device should be used */
+	strncpy(device, drvthis->config_get_string(drvthis->name, "Device", 0, DEFAULT_DEVICE),
+		sizeof(device));
+	device[sizeof(device) - 1] = '\0';
```
haraldg, line 79: Is there any point in creating this copy? Can't we just directly fopen() the result from config_get_string() ?

server/drivers/linux_devlcd.c
```diff
@@ -0,0 +1,312 @@
...
+		}
+	}
+
+	// Allocate the framebuffer
+	p->framebuf = malloc(p->width * p->height);
+	if (p->framebuf == NULL) {
+		report(RPT_ERR, "%s: unable to create framebuffer", drvthis->name);
+		return -1;
+	}
+	memset(p->framebuf, ' ', p->width * p->height);
```
haraldg, line 110: Do we even need a framebuffer for this type of device? I would guess (too lazy to check), that the kernel already has its own framebuffer.

I have no strong opinion, but I wonder if there is any implementation advantage from having our own?

Also you are creating a memory leak here. (You return an error without freeing the private data allocated above.) IMO it is okay to treat the alloc functions as infallible nowadays. A process is more likely to be killed by the OOM-killer then malloc returning an error. But if you do treat it as infallible, then you should cut out all the error checking and reporting.
(I'd just cut out the pointless checks, but if you want to keep them for the sake of correctness, then you have to actually do it right. There is no middle ground. ;-) - Of course this is moot if we actually don't need the framebuffer.)

server/drivers/linux_devlcd.c
```diff
@@ -0,0 +1,312 @@
...
+
+	/* open the device... */
+	if (!strcmp(device,"-")) {
+		p->fd = stdout;
```
haraldg, line 114: Is this only a leftover from the text driver or is there a legitimate use case with this driver?

server/drivers/linux_devlcd.c
```diff
@@ -0,0 +1,312 @@
...
+MODULE_EXPORT void
+linuxDevLcd_flush (Driver *drvthis)
+{
+	PrivateData *p = drvthis->private_data;
+	char out[LCD_MAX_WIDTH];
+	int i;
+
+	fprintf(p->fd, "\e[H"); // cursor home
+	for (i = 0; i < p->height; i++) {
+		memcpy(out, p->framebuf + (i * p->width), p->width);
```
haraldg, line 211: Do we need some check for escape sequences here?

server/drivers/linux_devlcd.c
```diff
@@ -0,0 +1,312 @@
...
+	y--; x--;
+
+	if ((x >= 0) && (y >= 0) && (x < p->width) && (y < p->height))
+		p->framebuf[(y * p->width) + x] = c;
+}
+
+
+/**
+ * Change the display contrast.
+ * linux /dev/lcd driver does not support this, so we ignore it.
```
haraldg, line 268: Then just don't define it at all.

haraldg (changes requested):
Thanks! This looks mostly good to me.

However one big question:
I'd assume, that the kernel already knows about the display size. Is there any way for us to get this info from the kernel ABI instead of duplicating it in LCDd.conf?

I have looked into the links you provided and the answers there seems to be: Go read the kernel code yourself. :-( I don't have time for this right now. But if it is possible, I think you should do it that way.

server/drivers/linux_devlcd.c
penfold42: all the drivers seem to do this. It gets used in error/info logging later on

server/drivers/linux_devlcd.c
penfold42: it was, but I also used for debugging. not sure theres a credible production use case for it.
I'll nuke it

server/drivers/linux_devlcd.c
penfold42: ok

penfold42:
> However one big question: I'd assume, that the kernel already knows about the display size. Is there any way for us to get this info from the kernel ABI instead of duplicating it in LCDd.conf?

Sadly not that I can see - I've looked thru the kernel source of charlcd,hd44780 and hd44780_common and none of them expose anything. /dev/lcd is also write only and for completeness I went on a grepping spree through /proc and /sys and found nothing.

server/drivers/linux_devlcd.c
penfold42: Do you mean to filter out a user incorporating escape sequences in their strings causing odd effects ?
If so, the quickest (and dirtiest) way would be to drop any \e characters we see.
The more complete way is to re-implement the kernel's parser but that could be fiddly and need ongoing maintenance.
(If the kernel required the escape sequences to be ; terminated it would be much easier, but only one of them is ; terminated)

haraldg: If \e is the only character, that might cause trouble, then just replacing \e with some other inoffensive character would be the way to go.

Again this seems to boil down to reading the kernel code. Which is kind of bad because the kernel code might change without us noticing. Such things should be specified somewhere. But it's not your fault. As long as we do reasonable guarding on our end, we are good.

haraldg:
> > However one big question: I'd assume, that the kernel already knows about the display size. Is there any way for us to get this info from the kernel ABI instead of duplicating it in LCDd.conf?
> 
> Sadly not that I can see - I've looked thru the kernel source of charlcd,hd44780 and hd44780_common and none of them expose anything. /dev/lcd is also write only and for completeness I went on a grepping spree through /proc and /sys and found nothing.

Thanks for checking. It would have been in charlcd source. But the driver defines neither read nor ioctl operations, so we are out of luck. You don't feel like getting involved with kernel development, do you? ;-)

server/drivers/linux_devlcd.c
haraldg: If you reorder the code slightly, you can use the pointer from config_get_string for everything before it is invalidated by calling config_get_string again. Also looking into this again, I see that the RPT_INFO lines are redundant. I think, one of them should be dropped.

I usually assume but don't trust, that the people before me knew what they were doing. Usually it serves me quite well. (And yes, maybe I'm a bit more into green programming than other people.)

penfold42:
I've covered all these ( i think!? ) and also added support for customer characters which brings hbar, vbar, bignum etc...
At the moment its in a [separate branch](https://github.com/penfold42/lcdproc/tree/linux_devlcd2).
Do you want me to merge these commits with my original branch or would you prefer a PR ?

haraldg:
> I've covered all these ( i think!? ) and also added support for customer characters which brings hbar, vbar, bignum etc... At the moment its in a [separate branch](https://github.com/penfold42/lcdproc/tree/linux_devlcd2). Do you want me to merge these commits with my original branch or would you prefer a PR ?

This is quite a lot of changes. I think it is easiest to review it from scratch, so maybe make a new PR. In the future please push fixes related to review comments to the original branch and put unrelated changes like new features into a new branch / PR.

As for the changes themself: A lot of it seem duplication of the hd44780 driver, which is a problem for two reasons:
1. Code duplication is bad.
2. We can't assume that the device behind /dev/lcd actually is a hd44780, can we? So any hd44780 specific code might cause trouble. We should stick to the generic parts of the kernel interface and let the kernel do all the heavy lifting.

Yes, this means that the devlcd driver won't support all the features that the hd44780 driver does. IMO that's okay. And if anybody wants to improve this, they WILL need to get the kernel stuff in shape to allow querying supported features etc.

willfe:
@penfold42 (gentle ping)

How would you like to proceed with this PR? Have you given any thought to @haraldg's feedback?

penfold42:
I propose we close it.

willfe:
Roger that, closing as requested.

Issue #217, opened 2024-11-16
hd44780: gpiod driver uses legacy call


manio:
Hi,
It seems `gpiod_chip_open_lookup` is not available anymore since https://github.com/brgl/libgpiod/commit/eb690e46efd09cd9a758ed1b19f99601f7b5de8a

I've got this when trying to start the driver:
Could not open driver module /usr/src/lcdproc/server/drivers/hd44780.: /usr/src/lcdproc/server/drivers/hd44780.so: undefined symbol: gpiod_chip_open_lookup

cc: @haraldg

manio:
Here are another problems:
```
gcc -std=gnu99 -DHAVE_CONFIG_H -I. -I../..  -I../..  -fPIC -Wall  -O3 -Wno-unused-function -MT hd44780-hd44780-gpiod.o -MD -MP -MF .deps/hd44780-hd44780-gpiod.Tpo -c -o hd44780-hd44780-gpiod.o `test -f 'hd44780-gpiod.c' || echo './'`hd44780-gpiod.c
hd44780-gpiod.c: In function 'init_gpio_pin':
hd44780-gpiod.c:73:2: warning: implicit declaration of function 'gpiod_chip_get_line' [-Wimplicit-function-declaration]
  *pin = gpiod_chip_get_line(pins->chip, number);
  ^
hd44780-gpiod.c:73:7: warning: assignment makes pointer from integer without a cast
  *pin = gpiod_chip_get_line(pins->chip, number);
       ^
hd44780-gpiod.c:80:2: warning: implicit declaration of function 'gpiod_line_request_output' [-Wimplicit-function-declaration]
  if (gpiod_line_request_output(*pin, "LCDd", 0) < 0) {
  ^
hd44780-gpiod.c: In function 'send_nibble':
hd44780-gpiod.c:104:2: warning: implicit declaration of function 'gpiod_line_set_value' [-Wimplicit-function-declaration]
  gpiod_line_set_value(pins->d7, ch & 0x08);
  ^
hd44780-gpiod.c: In function 'hd_init_gpiod':
hd44780-gpiod.c:150:2: warning: implicit declaration of function 'gpiod_chip_open_lookup' [-Wimplicit-function-declaration]
  pins->chip = gpiod_chip_open_lookup(chip);
  ^
hd44780-gpiod.c:150:13: warning: assignment makes pointer from integer without a cast
  pins->chip = gpiod_chip_open_lookup(chip);
             ^
```

[...]

Issue #214, opened 2024-05-04
hd44780-gpiod fails to compile with libgpiod-2.1.1 and gcc-14


heitbaum:
With gcc13 the compile worked - but will have issues with the libraries - as these functions are not available in libgpiod2.

```
Making all in drivers
make[3]: Entering directory '/var/media/DATA/home-rudi/LibreELEC.tv/build.LibreELEC-Generic.x86_64-13.0-devel/build/lcdd-0e2ce9b9c46c47363436f9ee730f7c71bf455f0f/.x86_64-libreelec-linux-gnu/server/drivers'
[...]

PR #219, opened 2024-12-07
hd44780: Add support for libgpiod2

manio:
Hi,
The old libgpiod1x library is nowadays not available in new OS releases (it can't even be compiled on my old `raspbian`), so change the original @haraldg gpiod1x driver implementation to add a support for the newest libgpiod2 library and decent kernel implementations.

This is fixing #214 and #217.

If you're OK with the changes, please do normal merge and don't squash because I've split the changes into separate commits for easy reviewing in the future.

### DIFF
--- a/server/drivers/hd44780-gpiod.c
+++ b/server/drivers/hd44780-gpiod.c
@@ -36,17 +36,39 @@ void gpiod_HD44780_reset(PrivateData *p);
 void gpiod_HD44780_close(PrivateData *p);
 
 typedef struct {
+	uint offset;
+	struct gpiod_line_request *req;
+	struct gpiod_line_settings *settings;
+	struct gpiod_line_config *line_cfg;
+	struct gpiod_request_config *req_cfg;
+} gpio_pin;
+
+typedef struct {
 	struct gpiod_chip *chip;
-	struct gpiod_line *en;
-	struct gpiod_line *rs;
-	struct gpiod_line *d7;
-	struct gpiod_line *d6;
-	struct gpiod_line *d5;
-	struct gpiod_line *d4;
-	struct gpiod_line *en2;
-	struct gpiod_line *bl;
-	struct gpiod_line *rw;
+	gpio_pin *en;
+	gpio_pin *rs;
+	gpio_pin *d7;
+	gpio_pin *d6;
+	gpio_pin *d5;
+	gpio_pin *d4;
+	gpio_pin *en2;
+	gpio_pin *bl;
+	gpio_pin *rw;
 } gpio_pins;
+
+/**
+ * Set GPIO pin state.
+ *
+ * \param pin     Pointer to gpio_pin structure.
+ * \param value   The new PIN state to set.
+ */
+static int
+gpio_set_value(gpio_pin *pin, int value)
+{
+	enum gpiod_line_value new_val = value ? GPIOD_LINE_VALUE_ACTIVE : GPIOD_LINE_VALUE_INACTIVE;
+	return gpiod_line_request_set_value(pin->req, pin->offset, new_val);
+}
+
 
 /**
  * Initialize a struct gpiod_line context by reading the related configuration
@@ -58,11 +80,11 @@ typedef struct {
  * \return          0 on success; -1 on error.
  */
 static int
-init_gpio_pin(Driver *drvthis, struct gpiod_line **pin, const char *name)
+init_gpio_pin(Driver *drvthis, gpio_pin **gpin, const char *name)
 {
 	PrivateData *p = (PrivateData *) drvthis->private_data;
 	gpio_pins *pins = (gpio_pins *) p->connection_data;
-	char config_key[8];
+	char config_key[10];
 	int number;
 
 	snprintf(config_key, sizeof(config_key), "pin_%s", name);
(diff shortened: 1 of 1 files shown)


manio:
Just a side note: I was using my raspberry-connected LCD display using:
`ConnectionType=raspberrypi`
and it was working fine just after start, but then later it always ended up with a garbage (I was trying with `RefreshDisplay=`, `DelayMult=`, etc. but doesn't help).

Now when using my new libgpiod2 implementation, it is working like a charm! :)

haraldg:
Thanks for looking into this!

First question: Does this code work with the old and new version of libgpiod or only the new one?

Rant: I'm quite annoyed, that libgpiod broke the API without a way to have both versions coexist (like libusb did) and even without a standardized (future proof) way to check the installed version. What a mess.

> please do normal merge and don't squash because I've split the changes into separate commits for easy reviewing in the future.

Actually your way to split the changes into commits makes it much harder to review them, because they are not independent. After spending about 30 Minutes switching back and forth between commits and cross-referencing things, I gave up.

As a guideline every commit should be useful on its own. At the very least the code should compile. Actually every commit should be compile tested.

Things like renaming arguments without changing the function body are a no go.

Please resubmit.

manio:
> Thanks for looking into this!
> 
> First question: Does this code work with the old and new version of libgpiod or only the new one?

only the new one

> Rant: I'm quite annoyed, that libgpiod broke the API without a way to have both versions coexist (like libusb did) and even without a standardized (future proof) way to check the installed version. What a mess.
> 
> > please do normal merge and don't squash because I've split the changes into separate commits for easy reviewing in the future.
> 
> Actually your way to split the changes into commits makes it much harder to review them, because they are not independent. After spending about 30 Minutes switching back and forth between commits and cross-referencing things, I gave up.

you could just go to https://github.com/lcdproc/lcdproc/pull/219/files and you'd have all changes on one screen, and if you do it locally you could do `git diff` with range. Or just a `git diff` on a whole merge commit.

> As a guideline every commit should be useful on its own. At the very least the code should compile. Actually every commit should be compile tested.

I spent a lot of time committing the changes as logical steps instead of one big "blob". I know it is not compiling on every step but this is just what merge commits are for. This is just a way I am doing my work, but OK - no problem - I am adjusting to your approach: I've squashed this for you and forced-pushed.

> Things like renaming arguments without changing the function body are a no go.

Sorry - what you mean by this? Can you do an inline-comment review?

haraldg:
> > Things like renaming arguments without changing the function body are a no go.
> 
> Sorry - what you mean by this? Can you do an inline-comment review?

In your old commit 0afa686c67187005ab0ad9571e63cb4d32e3bce2 you changed the argument to `init_gpio_pin()` from `pin` to `gpin` but you didn't make the same change in the body of the function. (You did it in a later commit.)

Your idea of logical commits is sound. I like to work like that myself and demand it, when reviewing bigger changes. However you didn't split at the right places.

If I have to use `git diff` on a range of commits during review, then clearly splitting wasn't helpful in the first place, was it?

server/drivers/hd44780-gpiod.c
```diff
@@ -35,19 +35,41 @@ void gpiod_HD44780_backlight(PrivateData *p, unsigned char state);
...
 } gpio_pins;
 
+/**
+ * Set GPIO pin state.
+ *
+ * \param pin     Pointer to gpio_pin structure.
+ * \param value   The new PIN state to set.
+ */
+static int
+gpiod_line_set_value(gpio_pin *pin, int value)
```
haraldg: Adding a local function with the same name as an (old) API function (or even just in the libgpiod "namespace") is confusing. I can see us do this, if you do a version check and write code, that supports both versions.

But otherwise I'd rename this function. Personally I'd just drop it and use the proper calls directly. They are not that complex.

server/drivers/hd44780-gpiod.c
```diff
@@ -58,29 +80,39 @@ typedef struct {
...
 		return -1;
 
-	*pin = gpiod_chip_get_line(pins->chip, number);
-	if (*pin == NULL) {
-		report(RPT_ERR, "init_gpio_pin: unable to request GPIO%d: %s",
-		       number, strerror(errno));
-		return -1;
-	}
+	*gpin = calloc(1, sizeof(gpio_pin));
+	gpio_pin *pin = *gpin;
```
haraldg: IIRC in body declarations are not LCDproc coding style.

Anyway, the naming convention between `pin` and `gpin`seems somewhat unintuitive. Probably the easiest to understand solution would be, to just add the asterisk each time to de-reference the double pointer.

If instead you want to change the function to return a pointer directly instead of passing the double pointer, that would also be okay.

(Yes, nitpicking that one. I wouldn't bother, if other changes weren't needed.)

server/drivers/hd44780-gpiod.c
```diff
@@ -90,6 +122,28 @@ init_gpio_pin(Driver *drvthis, struct gpiod_line **pin, const char *name)
 }
 
 
+/**
+ * Free related pin resources which should be freed.
+ *
+ * \param pin     Pointer to gpio_pin structure.
+ */
+void
+release_gpio_pin(gpio_pin *pin)
```
haraldg, line 130: declaration as `static` missing

server/drivers/hd44780-gpiod.c
```diff
@@ -90,6 +122,28 @@ init_gpio_pin(Driver *drvthis, struct gpiod_line **pin, const char *name)
...
+
+	if (pin->req_cfg)
+		gpiod_request_config_free(pin->req_cfg);
+	if (pin->line_cfg)
+		gpiod_line_config_free(pin->line_cfg);
+	if (pin->settings)
+		gpiod_line_settings_free(pin->settings);
+	if (pin->req)
+		gpiod_line_request_release(pin->req);
+}
```
haraldg, line 144: I think, you forgot to free `pin` here.

server/drivers/hd44780-gpiod.c
```diff
@@ -145,9 +200,10 @@ hd_init_gpiod(Driver *drvthis)
 		return -1;
 	}
 
-	pins = malloc(sizeof(gpio_pins));
+	pins = calloc(1, sizeof(gpio_pins));
 
-	pins->chip = gpiod_chip_open_lookup(chip);
+	snprintf(dev_path, sizeof(dev_path), "/dev/gpiochip%s", chip);
+	pins->chip = gpiod_chip_open(dev_path);
```
haraldg: I think, this changes behavior. But it wasn't explicitly documented behavior, so I hope it is okay.

It's a bit unfortunate though. The old call allowed to pass some unique identifier like `gpio.085f3b00`, that at least some systems have. If we are now back to sequentially indexed devices, then unstable probing order at boot time might actually break things.

haraldg (changes requested):

server/drivers/hd44780-gpiod.c
manio: I'll rather prefer to choose option to rename that function, otherwise we would end up with calls like:
`gpiod_line_request_set_value(pin->req, pin->offset, (state == BACKLIGHT_ON) ? GPIOD_LINE_VALUE_ACTIVE : GPIOD_LINE_VALUE_INACTIVE);`

server/drivers/hd44780-gpiod.c
manio: Yes, I was also thinking how to handle it when I was writing that code. But as you said it is currently undocumented, so I've changed the approach and I think the most flexible for user is to let this be a full path of a gpio device. User may also want to use eg. some UDEV rules (something similar like `/dev/disk/by-id...` - so restricting to `/dev/SOMETHING` is maybe not a good choice).

manio:
OK... pushed, all your code requests addressed.

haraldg:
I just added some minor updates to the documentation and committed this PR in https://github.com/lcdproc/lcdproc/commit/71877ee059a238400b5f6b41cee7c43a3df00334

manio:
Thank you!

